PR #37, opened 2026-02-16
add markdown format

### LINKED ISSUE #25: HTML to Markdown Support

matrv:
Closes #25

add HTML <-> Markdown and HTML <-> Markdown convention with the packages from the [unified ecosystem](https://unifiedjs.com/). I had to add a lot of npm packages, but they are all really small and do one small thing.

### DIFF
--- a/src/handlers/markdown.ts
+++ b/src/handlers/markdown.ts
@@ -1,0 +1,92 @@
+import type { FileData, FileFormat, FormatHandler } from "../FormatHandler.ts";
+import { unified, type Processor } from "unified";
+import remarkParse from "remark-parse";
+import remarkRehype from "remark-rehype";
+import rehypeStringify from "rehype-stringify";
+import rehypeParse from "rehype-parse";
+import rehypeRemark from "rehype-remark";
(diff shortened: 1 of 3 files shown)


src/handlers/markdown.ts
```diff
@@ -0,0 +1,92 @@
...
+  public ready = false;
+
+  private htmlToMarkdownPipeline?: Processor<any, any, any, any, string>;
+  private markdownToHtmlPipeline?: Processor<any, any, any, any, string>;
+
+  async init() {
+    this.markdownToHtmlPipeline = unified()
+      .use(remarkParse)
+      .use(remarkGfm)
+      .use(remarkRehype, { allowDangerousHtml: true })
```
flushedface: Why is `{ allowDangerousHtml: true }` set to true ?

matrv: You're right, it shouldn't be enabled. Just removed it

p2r3:
Type conventions are a bit weird here, but admittedly I don't really know the library. Otherwise, LGTM.

PR #60, opened 2026-02-16
Add dark mode

joan-code6:
Adding Darkmode
grafik


goog00goog:
yeah

p2r3:
Sorry, I think we should instead focus on a proper UI redesign rather than cramming features on top of this one. There's been a few proposals, feel free to propose a dark mode once we settle on one of those.

Issue #46, opened 2026-02-16
Add SQLite database format (.db)


piotrstenke:
Hey, I suggest adding support for sqlite databases. SQL tables can be pretty easily converted to a xlsx file, and since we already support xlsx, this should be straight-forward to implement (famous last words).

p2r3:
I don't think converting straight to XLSX is a good idea. Excel documents are very hard to construct. An easier alternative would be a simple HTML/SVG table. Or CSV, and then a separate handler for CSV to HTML.

PR #80, opened 2026-02-16
Add Sqlite3

forester302:
Add support for sqlite to csv

https://github.com/p2r3/convert/issues/46


p2r3:
This fails to build. Looks like you're including a JS library as TS? Did you forget to add the types package?

(You can check if it builds locally by running `npm run build` or `bun run build`, depending on your runtime.)

forester302:
should be good now.

PR #83, opened 2026-02-16
Add VTF input support and MIME normalization

Rhgx:
## Summary
- add a dedicated `vtf` format handler for Valve Texture Format input
- decode common VTF pixel formats (including DXT1/DXT3/DXT5 and common uncompressed layouts) to RGBA
- register `.vtf` in the handler registry and normalize `image/vtf` to `image/x-vtf`

## Validation
- `npm run build`

### DIFF
--- a/src/handlers/index.ts
+++ b/src/handlers/index.ts
@@ -12,6 +12,7 @@ import qoiFuHandler from "./qoi-fu.ts";
 import sppdHandler from "./sppd.ts";
 import threejsHandler from "./threejs.ts";
 import markdownHandler from "./markdown.ts";
+import vtfHandler from "./vtf.ts";
 
 const handlers: FormatHandler[] = [];
 try { handlers.push(new canvasToBlobHandler()) } catch (_) { };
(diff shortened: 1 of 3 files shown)


p2r3:
Where's this code from?

Rhgx:
I had a personal project I wrote a while back to convert VTFs into PNGs in JS using the [wiki ](https://developer.valvesoftware.com/wiki/VTF_(Valve_Texture_Format)) as a reference out of boredom. Asked GPT-Codex to convert it over into TS.

uriel-4:
It feels like for adding and merging support for more formats, it might be useful to include test files to check conversion and give more context as to what files were tested/supported and which use cases might have been broken after merge.
Also good for working with AI generated code, like this one, to always test it. Can be even part of GitHub actions on every PR merge.
Not really related to this specific PR, sorry, just a thought.

Rhgx:
@uriel-4 I have tested multiple VTFs from TF2 before posting the PR so I know the code works and also compared it to my original code to see if it made any mistakes.

p2r3:
> Can be even part of GitHub actions on every PR merge.

@uriel-4 We already run an action on every PR to check that the project builds. This ensures that even if the added code is faulty, it doesn't mess with the existing site. There's way too many file formats for us to individually vet every single contribution.

Issue #87, opened 2026-02-16
Allow conversion from any format


piotrstenke:
There should be an option to support conversion regardless of the source format. This would be useful for e.g. a ZIP converter (from any file to a ZIP). Without this feature, all supported types must be added to 'supportedFormats' either manually or at runtime, somehow.

PR #92, opened 2026-02-16
Add support for conversion from any format

piotrstenke:
#87

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,12 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "name": "Debug Vite Client",
+      "type": "chrome",
+      "request": "launch",
+      "url": "http://localhost:5173",
+      "webRoot": "${workspaceFolder}"
+    }
+  ]
+}
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ For a semi-technical overview of this tool, check out the video: https://youtu.b
 
 Here's how to deploy this project locally:
 
-1. Clone this repository.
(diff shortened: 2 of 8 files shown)


p2r3:
How did you test this?

piotrstenke:
> How did you test this?

I'm making a ZIP converter mentioned in #95, so I've tested with that, but a simple example would like this:

_anyHandler.ts_

```typescript
import type { FileData, FileFormat, FormatHandler } from "../FormatHandler.ts";

class anyHandler implements FormatHandler {
  public name: string = "any";

  public supportedFormats: FileFormat[] = [
    {
      name: "ZIP Archive",
      format: "zip",
      extension: "zip",
      mime: "application/zip",
      from: false,
      to: true,
      internal: "zip"
    }
  ];

  public supportAnyInput: boolean = true;

  public ready: boolean = false;

  async init() {
    this.ready = true;
  }

  async doConvert(
    inputFiles: FileData[],
    inputFormat: FileFormat,
    outputFormat: FileFormat
  ): Promise<FileData[]> {
    return [{ name: "test.zip", bytes: new Uint8Array([0, 1]) }];
  }
}

export default anyHandler;
```
_index.ts_

```typescript
try { handlers.push(new anyHandler()); } catch (_) { };
```

If you try to convert to ZIP from format listed in 'renameHandler' (e.g. xlsx), the 'renameHandler' will be used as usual. If you try with e.g. a png file, no direct path is found, so it fallbacks to this 'anyHandler'.
 
To test this you must temporarily comment out line 148 in _rename.ts_, as right now it throws an error for non-zip files.

```typescript
if (inputFormat.internal !== "zip") throw "Invalid input format.";
```

PR #94, opened 2026-02-16
Add FLO audio format.

mpopa4096-source:
My code is a boilerplate and doesn't have the encoding and decoding functions!
It only adds FLO as a valid format.
But you can take from https://www.npmjs.com/package/@flo-audio/libflo-audio

### DIFF
--- a/src/libflo.ts
+++ b/src/libflo.ts
@@ -1,0 +1,28 @@
+import type { FileData, FileFormat, FormatHandler } from "../FormatHandler.ts";
+
+import { Decoder, Encoder } from "libflo-audio";
+
+class libFloHandler implements FormatHandler {
+
+  public name: string = "libflo-audio";
+  public supportedFormats: FileFormat[] = [
+    {
+      name: "Waveform Audio File Format",
+      format: "wav",
+      extension: "wav",
+      mime: "audio/wav",
+      from: true,
+      to: true,
+      internal: "wav"
+    },
+    {
+      name: "flo™ Audio",
+      format: "flo",
+      extension: "flo",
+      mime: "audio/flo",
+      from: true,
+      to: true,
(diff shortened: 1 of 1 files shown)


NellowTCS:
Haha I wanted to add this as well! 👌 
(note: I created flo :D)

mpopa4096-source:
The goal here is to support ANYTHING! Anything that you can think of!
Yep, that's the goal of this project: You can add anything that you can think of!
Even MP3 to text using some ASR (Automatic Speech Recognition) model or text to MP3 using TTS (Text-to-Speech).

NellowTCS:
flo's mime type isn't `audio/flo` yet, as I still need to submit an application to IANA to claim it, however.
The web demo currently uses `application/octet-stream`, instead.

I think I can get decoding and encoding working, though! 
Using reflo would be better for conversion, however, as it even handles metadata: https://www.npmjs.com/package/@flo-audio/reflo

mpopa4096-source:
> flo's mime type isn't audio/flo yet, as I still need to submit an application to IANA to claim it, however.

I did that as a UNOFFICIAL MIME type.
Something like audio/x-flo would be better? @NellowTCS

p2r3:
Is this a draft..? Will this be worked on? There really isn't much here, even as far as boilerplate goes. The added code snippet isn't even valid TS.

mpopa4096-source:
> Is this a draft..? Will this be worked on? There really isn't much here, even as far as boilerplate goes. The added code snippet isn't even valid TS.

I mean, you should fix it. I'm not a expert! @p2r3

NellowTCS:
I was planning on doing this, before I saw the existing PR.

p2r3:
@NellowTCS It seems you're good to go make your own PR.

Issue #62, opened 2026-02-16
Add support for Minecraft Map Data


Peter-Maguire:
https://minecraft.wiki/w/Map_item_format  
  
It's an NBT container with a 128x128 paletted image inside, and there are plenty of examples of how to convert an image to a map on GitHub already.

JIYProgrammer:
When I saw this I worried that you suddenly invalidated my Minecraft world region format ticket. It relieves me greatly this is not so.

PR #127, opened 2026-02-16
Add Minecraft Map format

forester302:
https://github.com/p2r3/convert/issues/62
This is only the map to images

### DIFF
--- a/src/handlers/mcmap.ts
+++ b/src/handlers/mcmap.ts
@@ -170,7 +170,7 @@ function color_id_to_rgb(id: number): number[] | null {
     else if (shade_id == 2) shade_mul = 255
     else if (shade_id == 3) shade_mul = 135
 
-    return [Math.floor((r * shade_mul) / 255), Math.floor((r * shade_mul) / 255), Math.floor((r * shade_mul) / 255)];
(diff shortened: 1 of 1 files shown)


src/handlers/mcmap.ts
```diff
@@ -0,0 +1,176 @@
...
+                mime: "image/x-rgb",
+                from: false,
+                to: true,
+                internal: "rgb"
+            },
+            {
+                name: "Minecraft Map File",
+                format: "mcmap",
+                extension: "dat",
+                mime: "application/octet-stream", // I am required to put something here
```
p2r3: While there is obviously no MIME type for Minecraft maps, it's still preferable to invent a nonstandard MIME type (e.g. `application/x-minecraft-level`).

src/handlers/mcmap.ts
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    async doConvert(
+        inputFiles: FileData[],
+        inputFormat: FileFormat,
+        outputFormat: FileFormat
+    ): Promise<FileData[]> {
+        const outputFiles: FileData[] = [];
+
+        console.log(inputFiles[0].bytes, inputFormat, outputFormat)
```
p2r3: Stray console.log left over from development?

PR #73, opened 2026-02-16
Draft: UI Revamp

Jor02:
This PR introduces a major UI revamp of the converter application application.

## Progress
- [x] Create initial low-fidelity wireframes
- [x] Turn low-fidelity wireframes into high-fidelity ones (using Figma).
- [x] Turn high-fidelity wireframes into actual TSX + CSS.
- [x] Add dark/light theme toggle, make colors reactive
- [x] Split big TSX files into smaller reusable and maintainable components.
- [x] Cleanup main.ts for better maintainability 
- [x] Port all functionality of the existing application to the new buttons/components.

Anyone is free to provide any feedback on the design or implementation (even small nitpicks are appreciated) as we want to develop the best possible experience for users!

## Current version:
image

### Initial draft wireframe:
wireframe

This PR should:
- resolve #138 
- resolve #166 
- resolve #235
- resolve #310
- resolve #419
- resolve #21

PR #129, opened 2026-02-16
Refactor index.html for improved structure and semantics

o7528328-arch:


p2r3:
Rather unnecessary, especially considering #73.

PR #133, opened 2026-02-16
add txt to ps1, and ps1 to txt

1TSnakers:
just creates a simple script that prints what the txt file wrote, also goes backwards

### DIFF
--- a/src/handlers/index.ts
+++ b/src/handlers/index.ts
@@ -1,6 +1,8 @@
 import type { FormatHandler } from "../FormatHandler.ts";
 
 import canvasToBlobHandler from "./canvasToBlob.ts";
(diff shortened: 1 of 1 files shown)


1TSnakers:
the ps1 instructions do work, running the file doesnt (idk how to make it work):
image

p2r3:
A couple of thoughts. Firstly, the input should be sanitized to prevent escaping the string. Shell scripts are no joke. Secondly, you shouldn't need two handlers to handle input/output. Just one is enough, as long as you check what format is the input and what format is the output.

p2r3:
Marking this as draft until the issues get resolved.

1TSnakers:
someone did this with batch files, and their implantation is better

PR #135, opened 2026-02-16
Add a landing page to prevent loading 180+ MB as soon as page is loaded.

mountain658:
# Previously using tons of bandwidth:
image

Now it will have a very simple landing page and link to the repo.

(Good at least before we get the UI upgrade)

# Preview
image

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,52 +1,89 @@
 <!DOCTYPE html>
 <html lang="en">
+
 <head>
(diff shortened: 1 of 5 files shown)


p2r3:
Did you test the bandwidth use with caches built or not? It does not download all tools right away, but rather only when they're used. Test the deployed site.

PR #144, opened 2026-02-17
Fixes to audio > image > audio pipeline in meyda.ts

Basiliotornado:
Correct the amplitude calculation for audio > image by dividing result by bufferSize, and remove the windowing function in image > audio. 

Image before:
ChaozFantasy

Image after:
ChaozFantasy(1)
don't worry about the second image having little green, you could normalize based on the maximum amplitude if you prefer that method, i think this sounds fine

There's still some issues, there seems to be a cap at 15khz, and some stray green pixels show up that you can hear when converting back to audio, not sure about that.

### DIFF
--- a/src/handlers/meyda.ts
+++ b/src/handlers/meyda.ts
@@ -148,12 +148,6 @@ class meydaHandler implements FormatHandler {
         const sampleRate = this.#audioContext.sampleRate;
 
         const audioData = new Float32Array(imageWidth * hopSize + bufferSize);
-        const window = new Float32Array(bufferSize);
-
-        // Generate Hanning window
-        for (let i = 0; i < bufferSize; i++) {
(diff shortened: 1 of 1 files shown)


src/handlers/meyda.ts
```diff
@@ -244,7 +238,8 @@ class meydaHandler implements FormatHandler {
 
           const pixels = new Uint8ClampedArray(imageHeight * 4);
           for (let j = 0; j < imageHeight; j ++) {
-            const magnitude = Math.sqrt(real[j] * real[j] + imaginary[j] * imaginary[j]);
+            // Calculate amplitude, amplitude is halved when only half of the FFT is used, so double it
```
Basiliotornado, line 241: Realizing that maybe the comment isn't entirely necessary

p2r3:
YES! OMG! This bug has been haunting me!!! Thank you for spotting and fixing it. This is _so so so_ much cleaner.

Issue #70, opened 2026-02-16
Implement Pandoc for documents and markdown


pauliesnug:
Instead of the dedicated Open/MS office conversion logic, Pandoc has WASM available. https://pandoc.org/ https://github.com/haskell-wasm/pandoc-wasm

p2r3:
I've only just learned about Pandoc, but from what I can tell, it doesn't support spreadsheets/presentations, so not a drop-in replacement for Envelope. Still, it might be useful for all the other stuff that it does support. Not sure if it persists document formatting either - that's pretty important here.

pauliesnug:
I think it does support PPTX bidirectionally and converting from Microsoft Excel, at least based on the website.

p2r3:
Sounds enticing. We really do need a better solution for documents, it's kind of all over the place right now.

I think I'll put on hold any proposals for markdown-esque formats for now until someone starts work on implementing Pandoc. Seems to be the right way forward.

PR #153, opened 2026-02-17
Add support for more TXT encodings

millennIumAMbiguity:
BOM is an optional header in .txt files.

I was surprised to see that utf16 is not standardized as a supported format, and Chrome defaults to utf8 with BOM.
It was just last week that I last used a hex editor to write BOM to a .txt file

Added the following formats:
 - txt-utf8 (forced with BOM)
 - txt-utf8 (forced without BOM)
 - txt-utf16le
 - txt-utf16be
 - txt-utf32le
 - txt-utf32be
 
 All can be converted into and from `text/plain` *(meaning utf8 with or without BOM, depending on browser)*
 utf8 formats with forced BOM behavior are out only, as all web browsers support both variants.


p2r3:
Putting this on the backburner until #70 gets resolved

millennIumAMbiguity:
> Putting this on the backburner until #70 gets resolved

any update?

PR #168, opened 2026-02-17
Add VTF input support and MIME normalization

Rhgx:


Rhgx:
Was trying to delete the branch, my bad.

Issue #126, opened 2026-02-16
Converting from Video (mp4) to swf should include both video and audio


LauraWebdev:
Currently, converting from a video format such as mp4 usually takes a route through either gif or mp3 to convert to swf. This means depending on the path taken, the output swf either has the audio OR the video.

Examples:
mp4 -> gif -> swf (This swf lacks the audio present in the mp4)
mp4 -> mp3 -> swf (This swf lacks the video present in the mp4)

Sometimes, the path will also be taken through png apparently? Ending up with an swf of a still image of the visual interpretation of the audio.

What I would expect would be a path/conversion that includes both the video AND the audio, akin to swf content commonly found on old websites like z0r.de

PR #171, opened 2026-02-17
Fix conversion routing for mp4 -> swf

Rhgx:
Fixes [#126](https://github.com/p2r3/convert/issues/126).

This adjusts conversion path ordering so `mp4 -> swf` no longer prefers lossy intermediate routes like `gif` or `mp3` when a video-preserving route is available.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -314,6 +314,43 @@ const convertPathCache: Array<{
   node: ConvertPathNode
 }> = [];
 
+const videoContainerFormats = new Set([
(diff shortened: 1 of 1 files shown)


p2r3:
Hm, not really a fan of hardcoding format names, especially not in the algorithm part of the program. The MIME group stuff is fine, and probably sufficient? Individual formats can be sorted on the handler, like they already are for FFmpeg or ImageMagick, just adjusted for fidelity over compatibility.

mmewtwosaysbye:
@p2r3 just wondering why would this not be a good idea? seems okay to me.

p2r3:
Superseded by #230.

PR #172, opened 2026-02-17
Add support for Ableton Live Set files

robhanlon22:
.als files are just gzipped XML files 🎶


p2r3:
I'm not sure that this counts as properly supporting the format. This doesn't unfold the media inside the file, so at best this is just a decompression tool. The output is just as meaningless as the input for most use cases.

robhanlon22:
unfortunately i don't think that converting the embedded media to other formats is particularly feasible - the ableton format is absurdly flexible but it's also non-hermetic - lots of things are stored on-disk separate from the .als file because otherwise it would bloat the .als with tons of audio data. some audio data is stored in the file — for instance, things recorded via the Looper device are embedded and portable, but i think the "real" solution would be uploading a compressed ableton project directory, identifying that it's an ableton project, then doing the magic there. most definitely a heavier lift and since ableton project directories can tend toward multi-gigabyte sizes maybe a bit untenable. would be interested in your thoughts though 😄

PR #181, opened 2026-02-17
feat: implement a custom bfs search and visualisation

Relst:
# Note:
- It would be great if this part could be validated as if this is wrong then the entire visualisation is wrong, as the key assumption I made is that the handlers are universal converters i.e. can convert any input to any output specified by from and to [Cartesian Product of FROM and TO formats of every handler or more simply every pair of FROM and TO].
- As there was only a boolean from and to in the search algorithm, on which I could base this visualisation. A very strong assumption.
- This is more of an idea, rather than a full on implementation as all of the nodes are embedded directly into the html by pyvis, and as I didn't want to add even more files which contained the node data I left it as so.

## Description
This PR introduces a dedicated visualisation tool and custom BFS to map conversion paths . It provides a visual representation of how files move through various handlers, although it is not a 1-1 due to no semantics of loss-less conversion, it does serve as both a visualisation guide and a different representation which could be faster if semantics of loss-less and priority are introduced (Dijkstra's, A* or Floyd–Warshall algorithm).

- **Dual-Mode Exploration:** 
   - **Basic Mode:** Uses MIME types for a high-level overview of media transitions.
  - **Advanced Mode:** Uses specific format identifiers (e.g., `r3d`, `jpeg`) for granular debugging of handler I/O.
- **Strict Implementation:** Due to the implementation of the graph, and possibly my assumption being wrong we ensure within the JS that the path we find is an actual valid conversion path from i.e. png -> jpg, has an actual conversion in a given handler.

## Use Case & Benefits
- **Development :**  Highlights issues in the current system, as the version used within the converter sometimes takes forever to find a valid path, and suggests system might benefit from moving toward other pathfinding algorithms and representations.
- **Expansion:** As the number of formats increases, serves as a new way to debug and visualise changes.
- **Cool:** It's just a cool visualisation.

## Methodology
- The visualisation utilizes a validation map consisting of `HANDLER`, `INPUT_FORMAT`, `INPUT_MIME`, `OUTPUT_FORMAT`, and `OUTPUT_MIME`. 
- Where we iterate over every row to get the semantic mapping of Format -> Handler  and Handler ->  Formats, and then generate the full graph.
- An example row is : "canvasToBlob","png","image/png","jpeg","image/jpeg"

## Validation
To verify the accuracy of the logic, I tested against "Ghost Paths" I generated during development. 
- **Example:** A path from `png -> svg` is valid, and `jpg -> html` is valid. However, if a handler supports both but not the cross-conversion (e.g., `png -> html`), the tool correctly identifies this as an invalid path due to the lack of directed connections in the sparse representation (Not an issue for now due to our assumption).
- This ensures that if a user needs to convert Format A to Format X, the visualization confirms the handler in between is explicitly capable of that specific transition.
- This can be seen using the from_png and from_jpeg, and to_svg and to_html, in the disconnected component with converter GHOST_HUNTER

## Visualisation
Below showcases me attempting to convert an r3d file to an svgz. However the search takes for ever to find a valid path, and thus an easier path is discovered by my graph representation of conversions.
Screenshot 2026-02-17 at 9 27 56 PM
Screenshot 2026-02-17 at 8 27 14 PM
Screenshot 2026-02-17 at 8 27 53 PM
Screenshot 2026-02-17 at 8 28 59 PM

## Disclaimer
- The full automated metadata extraction pipeline is not included in this PR; it currently relies on the provided format/handler mappings, and my current implementation for extracting data is not great :(, and I am unsure even if my assumption is correct.
- Parts of the UI and Physics optimisation were refined with AI assistance to improve performance.
- The tool currently priorities **reachability** over **loss-less conversion** and I have not validated a conversion of an actual r3d file using that format from the photos above, as r3d files are massive and downloading one from the samples, requires an account which I'm currently not up for.
- Ideally this serves as an initial guide which can be further maintained and explored with clarifications from the owner.
- Even if a path is shown on the graph it is not necessary the converter will even be able to convert between the valid methods i.e. trying to convert from html -> gif87, has a path html -> svg -> gif87, however svg -> gif87, does not work or takes too long to find a path.

Screenshot 2026-02-17 at 10 01 32 PM

## For getting Handlers CSV
- I have had Gemini generate this as the original one I used has already been deleted, and this generates everything mine had + a couple extra from the newer additions. 
- If we aren't able to have a cartesian product of FROM and TO, then all that is really required is to update the FROM and TO collection, and generating the graph is pretty trivial. However as previously stated from my reading of the code in main.ts, it does appear that we make the cartesian product assumption.
```
for (const handler of handlers) {
    const formats = window.supportedFormatCache.get(handler.name);
    if (!formats) {
      console.warn(`No cache found for ${handler.name}. Try initialising it first.`);
      continue;
    }

    const readers = formats.filter(f => f.from);
    const writers = formats.filter(f => f.to);

    for (const input of readers) {
      for (const output of writers) {
        // Skip same-format identity conversions
        if (input.format === output.format && input.mime === output.mime) continue;

        rows.push([
          handler.name,
          input.format,
          input.mime,
          output.format,
          output.mime
        ]);
      }
    }
  }
```
---


p2r3:
I feel like this is vastly outside of the scope of this repository. Having to upkeep a fancy visualizer should not be one of the requirements for a file converter.

p2r3:
Obsoleted by #230

PR #183, opened 2026-02-17
feat: add SVG tracer via imagetracer (fix #178)

### LINKED ISSUE #178: Converting PNG to SVG simply embed png as image, instead of converting
When converting a PNG image to SVG (using the `ImageMagickHandler` converter), it simply ends up with svg like this

```svg
<svg ..
<image id="image0" width="1117" height="843" x="0" y="0"
    xlink:href="data:image/png;base64,iVB...<PNG blob>"
```

Image viewers accept this, but due to SVG it being only a container, it can break conversions from svg to other formats.
(case: I try to write a svg -> python turtle converter, and I want png -> svg -> pyTurtle flow to work )

It would be nice if p2r3convert could [trace](https://en.wikipedia.org/wiki/Image_tracing) images to svg. 

https://github.com/murongg/imagetracer is the only js project I found which worked well.

* ImageMagick - Have basic edge detection feature, but cannot convert that into a colored svg
* node-potrace - managed to only trace half of the images
* Inkscape - The standard tool for tracing, [currently](https://gitlab.com/inkscape/inkscape/-/merge_requests/2032) no wasm/js port

matan-h:
This PR add support for tracing SVG (instead of embedding them, #178).
This also change the main path in simple mode to take the more specific handler if either can be used.

1. add handler to convert image -> svg via imagetracer module
2. add sort by specific to simple mode so more specific handlers are used by default


src/handlers/index.ts
p2r3, line 1: Stray newlines after the handler?

src/main.ts
```diff
@@ -389,6 +389,10 @@ async function buildConvertPath (
       const candidates = allOptions.filter(opt =>
         validHandlers.includes(opt.handler) &&
         opt.format.mime === target.format.mime && opt.format.to
+      ).sort((a,b)=> // always take the more specific handler (less output formats) if there is doubt
+            (a.handler.supportedFormats?.filter(f => f.to).length ?? 0) -
+            (b.handler.supportedFormats?.filter(f => f.to).length ?? 0)
+
```
p2r3: Changing the routing algorithm is outside of the scope of a pull request which aims to just add a conversion route.

You do not need to modify the algorithm for this. Just put your handler above canvasToBlob (which currently handles SVG to PNG) in `index.ts`, and it will be prioritized for the formats that it supports.

src/handlers/index.ts
```diff
@@ -6,6 +6,8 @@ import htmlEmbedHandler from "./htmlEmbed.ts";
 import FFmpegHandler from "./FFmpeg.ts";
 import pdftoimgHandler from "./pdftoimg.ts";
 import ImageMagickHandler from "./ImageMagick.ts";
+import svgTraceHandler from "./svgTrace.ts";
+import renameHandler from "./rename.ts";
```
p2r3: This seems to have been improperly merged - the rename handler has been split up into zip and text variants, see the line below.

In other words, this line should be removed.

PR #186, opened 2026-02-17
Add image to plaintext conversion

zipsegv:
This adds a conversion from images to text by rendering it out in ASCII, using [a simple library I created today](https://git.sr.ht/~thezipcreator/image-to-txt). For simplicity, this just modifies `canvasToBlobHandler` to add a plaintext output; this way image rendering logic does not need to be duplicated.

This leads to some... interesting... compound conversions (try converting a pdf to txt if you want to have some fun).

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,3 +7,6 @@
 [submodule "src/handlers/sppd"]
 	path = src/handlers/sppd
 	url = https://github.com/p2r3/sppd
+[submodule "src/handlers/image-to-txt"]
+	path = src/handlers/image-to-txt
+	url = https://git.sr.ht/~thezipcreator/image-to-txt
--- a/src/handlers/canvasToBlob.ts
+++ b/src/handlers/canvasToBlob.ts
@@ -1,4 +1,5 @@
 import type { FileData, FileFormat, FormatHandler } from "../FormatHandler.ts";
+import { imageToText, rgbaToGrayscale } from "./image-to-txt/src/convert.ts";
 
 class canvasToBlobHandler implements FormatHandler {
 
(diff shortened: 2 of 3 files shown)


p2r3:
Ooh, exciting 👀

Also potentially closes a few of the issues complaining about missing TXT file routing.

zipsegv:
> Also potentially closes a few of the issues complaining about missing TXT file routing.

imo there should probably be more "document > txt" type things, because if you try to convert any text document into plaintext and it goes through an image then all you get out is nonsense that vaguely looks like the original document, lol.

if it's possible, this route should probably be deprioritized as much as possible.

p2r3:
In the future, sure. When we have a proper markdown handler (*cough* Pandoc *cough*), that can be put higher up on the list.

GuilhermeRaposo:
@TheZipCreator The `rgbaToGrayscale` could be improved a bit:

```JS
export function rgbaToGrayscale(
    r: number,
    g: number,
    b: number,
    a: number,
): number {
    const rr = r * a + 1 * (1 - a);
    const gg = g * a + 1 * (1 - a);
    const bb = b * a + 1 * (1 - a);

    return 0.2126 * rr + 0.7152 * gg + 0.0722 * bb;
}
```

The original function produced wrong results for semi-transparent pixels because it multiplied each channel by alpha directly. I've also updated it to use BT.709 coefficients which is the modern standard on HD displays.

These two changes already made a small but noticeable difference in the output ASCII for me.

zipsegv:
@GuilhermeRaposo I've tried this out but I don't really like how it produces white for transparent pixels; I think it looks a lot better when the background is black (and therefore produces space for transparent pixels).

isn't multiplying by alpha equivalent to placing the image over a black background?

PR #188, opened 2026-02-17
feat: add export from svg to python turtle code

matan-h:
This PR add support for the conversion SVG -> python turtle code (not perfect conversion, due to turtle limitations)

note that it the path to convert something else to turtle (X->SVG->turtle) blocked by #178 

This image is from a generated ~11,000 LOC python file :)

image


p2r3:
> note that it the path to convert something else to turtle (X->SVG->turtle) blocked by 178

I'm not sure what you meant by this. Did you mean to say that this conversion doesn't work because it's blocked by the existing SVG-to-image handler? If so, moving your handler up on the list above canvasToBlob should fix that.

matan-h:
I meant that good SVG would convert perfectly to turtle.

If the user upload a *PNG file* and try to convert that into turtle code, then because the PNG->SVG conversion does not actually convert, it will convert into an empty turtle code.

p2r3:
Ah, okay. I suppose it makes sense to merge #183 first, then.

src/handlers/index.ts
```diff
@@ -17,6 +17,7 @@ import markdownHandler from "./markdown.ts";
 import vtfHandler from "./vtf.ts";
 import mcMapHandler from "./mcmap.ts";
 import jszipHandler from "./jszip.ts";
+import PyTurtleHandler from "./pyTurtle.ts";
```
p2r3: Please use camelCase for these imports/exports. That is, first letter lowercase, the rest as they already are.

matan-h: I didnt notice that, thanks

matan-h:
I now notice that since TXT got merged too we both use text/plain as mimetype, therefore when trying to convert to turtle, it output the TXT instead. I should probably change it to be `text/x-python`

PR #199, opened 2026-02-17
Create pull_request_template.md for PR submissions

alex-ljohnson:
Simple change to add a pull request template for contributors to follow.
Just a basic template for now
Can be additionally edited as seen fit once merged.


p2r3:
I think this is unnecessary at the moment. We haven't had many issues with malformed PRs, so having a template would just cause unnecessary friction. I might consider it if it becomes relevant in the future.

PR #200, opened 2026-02-17
Add plugin registration API for custom handlers

yosr481:
This PR adds a plugin registration API so you can add custom handlers without forking the repo (not a perfect hook, but zero-touch for core files)

Note that you can just drop files into src/plugins/ and they'll get picked up automatically at build time

This allows adding new formats without touching src/handlers/index.ts every single time :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,9 @@ The best way to contribute is by adding support for new file formats (duh). Here
 
 ### Creating a handler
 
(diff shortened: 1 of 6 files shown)


p2r3:
No.

`index.ts` serves a purpose. It lets us define an _order_ for imports, which _does_ matter, as some routes take priority over others.

This is also just a very noisy PR, touching and reformatting a bunch of stuff that it does not relate to.

PR #201, opened 2026-02-17
Windows batch to Win64 exe

millennIumAMbiguity:
Works by having the exe create a .bat in %temp% and runs it.
The file is deleted after the script finishes.

The .bat file is patched as binary data inside the exe and is written as a binary stream.
This way, any text encoding (that Windows supports) can be used.

It uses a fixed-length EXE template to avoid triggering modification validations.

### DIFF
--- a/src/handlers/batToExe.ts
+++ b/src/handlers/batToExe.ts
@@ -8,7 +8,12 @@ class batToExeHandler implements FormatHandler {
   ];
   public ready = false;
 
+  private header: Uint8Array|null = null;
+  private footer: Uint8Array|null = null;
+
   async init() {
+    this.header = await fetch("/src/handlers/batToExe/exe65824head.bin").then(res => res.arrayBuffer()).then(buf => new Uint8Array(buf));
+    this.footer = await fetch("/src/handlers/batToExe/exe65824foot.bin").then(res => res.arrayBuffer()).then(buf => new Uint8Array(buf));;
     this.ready = true;
   }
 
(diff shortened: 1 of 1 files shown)


p2r3:
Surely, at the very least, we could generate the "padding" in code?

I believe an even better solution would be to download this file as a binary blob in `init()`. This would ensure that it only gets loaded when the handler is used. Otherwise the user is downloading the whole stub binary every time they open the page.

You can put the binary in its own folder, under `/src/handlers/batToExe/___.exe` for example, and then just `fetch()` it from the TS code.

millennIumAMbiguity:
> Surely, at the very least, we could generate the "padding" in code?
> 
> I believe an even better solution would be to download this file as a binary blob in `init()`. This would ensure that it only gets loaded when the handler is used. Otherwise the user is downloading the whole stub binary every time they open the page.
> 
> You can put the binary in its own folder, under `/src/handlers/batToExe/___.exe` for example, and then just `fetch()` it from the TS code.

Do you mean like this?
I also split the exe into the header and footer binaries so we know the respective lengths.

zipsegv:
imo it would be nice if you also had instructions on how to generate the binaries somewhere in the repository, for reproducibility (if you compiled source code to generate them, that should be here probably).

src/handlers/batToExe.ts
```diff
@@ -16,48 +15,61 @@ class batToExeHandler implements FormatHandler {
...
+    outputFormat: FileFormat,
   ): Promise<FileData[]> {
+    // 0x0 -> 0x22CDF
+    const header = await (
+      await fetch("/src/handlers/batToExe/exe65824head.bin")
+    ).bytes();
+    // 0x32E00 -> 0x339EA
+    const footer = await (
+      await fetch("/src/handlers/batToExe/exe65824foot.bin")
+    ).bytes();
```
p2r3: I'm not sure that this is the correct place to `fetch` from? Our site lives under `/convert`, so I think the correct approach would be to fetch `/convert/src/handlers/batToExe/...`, or just `src/handlers/batToExe/...` without the leading slash.

millennIumAMbiguity: huh, but it worked in docker..
Oh, I see, it failed build after I rebased it. :(

PR #206, opened 2026-02-17
Introduce CommonFormats to reduce repeating code (#202)

### LINKED ISSUE #202: Reduce Duplicate Format Definitions Across Files
Hi,

While reviewing the code, I noticed that certain formats (for example PNG) are defined multiple times across different files, like this:

```js
{
  name: "Portable Network Graphics",
  format: "png",
  extension: "png",
  mime: "image/png",
  from: true,
  to: true,
  internal: "png"
}
```

Although I understand that there are differences between `from` and `to`, the core format definition itself appears to be duplicated.

I would like to suggest centralizing all format definitions in a single shared module, and then retrieving them through a helper function such as:
```js
getFormat(name, from, to)
```

BoBkiNN:
Resolves #202 
---
This PR adds a CommonFormats object and several related classes to reduce repeated format typing.

Now, instead of repeating same file formats extension, name, mimetype, helper method `supported` can be used on FormatDefinition instances. 
This method creates a `FileFormat` object from format definition which can be used by handlers as usual. 

This change do not break any old code.

I have introduced several common formats instances and replaced their usages with usage of CommonFormats

### DIFF
--- a/src/CommonFormats.ts
+++ b/src/CommonFormats.ts
@@ -54,6 +54,12 @@ const CommonFormats = {
         "html",
         "text/html"
     ),
+    MD: new FormatDefinition(
+        "Markdown Document",
+        "md",
+        "md",
+        "text/markdown"
+    ),
     // audio
     MP3: new FormatDefinition(
         "MP3 Audio",
--- a/src/handlers/markdown.ts
+++ b/src/handlers/markdown.ts
@@ -13,15 +13,7 @@ class markdownHandler implements FormatHandler {
 
   public name = "markdown";
   public supportedFormats = [
-    {
-      name: "Markdown Document",
-      format: "md",
-      extension: "md",
-      mime: "text/markdown",
-      from: true,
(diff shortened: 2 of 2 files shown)


p2r3:
Hm, yes, not a bad idea. Though I do think that this PR should then probably also update the contribution instructions in the readme, as well as adapt a few other common formats. PNG and JSON are fine as a boilerplate example, but opening a second PR just to fill in more formats seems redundant.

I might do that myself when I get the time, though if you wish, feel free to go through it yourself. I'm quite busy at the moment with the volume of other issues and all.

BoBkiNN:
I will replace some other common and popular formats a little bit later in this PR. Also updated readme contribution example

BoBkiNN:
Converted to draft to add more common formats

p2r3:
We can add more formats later - for now, I'll merge this PR as-is, as otherwise new format PRs are left waiting.

PR #207, opened 2026-02-17
fixed ffmpeg with and height error

### LINKED ISSUE #6: PDF to MP4 chooses ... interesting conversion path
When trying to convert a PDF to MP4, for some reason convert.to.it chooses the route PDF -> PNG -> WAV -> MP4. It appears that somehow it ignores that MP4 is supposed to be a video format (so PNG -> MP4 would be trivial with ffmpeg), and ends up with an MP4 audio file instead. When choosing MOV as the destination file format, it does pick PDF -> PNG -> MOV, so you can see this should in theory also work with MP4. but the resulting MOV file is not actually compatible with Quicktime on MacOS.

Timmindstorms:
Fixes #6 

For some video formats the with and height must be a multiple of X. When that error is throw the frame is padded to make it the correct hight and with.

### DIFF
--- a/src/handlers/FFmpeg.ts
+++ b/src/handlers/FFmpeg.ts
@@ -226,12 +226,19 @@ class FFmpegHandler implements FormatHandler {
 
     if (stdout.includes("Conversion failed!\n")) {
 
-      if (!args) {
-        if (stdout.includes("Valid sizes are")) {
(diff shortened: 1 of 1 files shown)


p2r3:
Thanks for debugging this!

PR #197, opened 2026-02-17
Add converters to and from JSON

zipsegv:
Specifically, this adds:

* JSON <-> CSV (conversion from JSON to CSV is kind of janky because CSV has no nesting but oh well)
* JSON <-> XML
* JSON <-> YAML

PR #208, opened 2026-02-17
fix action: missing npmrc in docker build

PoliceFighter761:
#197 introduced a .npmrc to add support for custom npm sources. That broke the docker build due to the file not being copied into the container. This PR corrects this behavior while also adding frozen-lockfile to ensure it behaves the same as the pages action.

Tested locally using the `convert:dev` tag

Action ran [successfully in my fork repository](https://github.com/PoliceFighter761/convert/actions/runs/22106634423/job/63891137231)


p2r3:
Awesome, thank you.

PR #213, opened 2026-02-17
Search algorithm optimizations & edge cases in FFMpeg and Meyda

mdev34-lab:
- Added validation to check if FFMpeg output exists before attempting to read
- ~~Added passthrough in Meyda in case input format is exactly the same as output format~~
- ~~Fixed path exploration to continue after conversion failures (idk)~~
- Added validation that the full conversion path was completed successfully
- Added BFS cycle protection (no more endless loops)
- Added dead end detection and more verbose logging for debug purposes

From testing this looks good to me locally, but maybe I broke something intended (please inform me if I did, I'm honestly not very confident in this PR).

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -422,7 +422,7 @@ async function buildConvertPath (
         if (path.some(c => c.format === format)) continue;
         
         const newPath = path.concat({ format, handler });
-        const pathSignature = newPath.map(c => c.format.format).join("→");
+        const pathSignature = newPath.map(c => c.format.format).join("•");
         
         console.log(`Exploring path: ${pathSignature}`);
         


src/handlers/meyda.ts
```diff
@@ -116,8 +116,12 @@ class meydaHandler implements FormatHandler {
...
     const hopSize = bufferSize / 2;
 
-    if (inputIsImage === outputIsImage) {
-      throw "Invalid input/output format.";
+    if (inputFormat.format === outputFormat.format && inputFormat.mime === outputFormat.mime) {
+      // Pass-through for identical formats
+      return inputFiles.map(file => ({
+        name: file.name.split(".")[0] + "." + outputFormat.extension,
+        bytes: file.bytes
+      }));
```
p2r3: This seems unnecessary - there's an explicit passthrough check in `main.ts`, and paths won't route through the exact same format twice anyway.

src/main.ts
```diff
@@ -344,10 +337,16 @@ async function attemptConvertPath (files: FileData[], path: ConvertPathNode[]) {
     } catch (e) {
       console.log(path.map(c => c.format.format));
       console.error(handler.name, `${path[i].format.format} → ${path[i + 1].format.format}`, e);
-      return null;
+      // Continue exploring other paths instead of returning null
```
p2r3: Uh, what? This doesn't make any sense.

The surrounding loop iterates through _steps_ of _one_ path, _not multiple paths_. Returning `null` here is an early exit to indicate that this one singular current path failed. `attemptConvertPath` is only given one path. It makes no sense to continue down a path where one link already failed, thus the return.

The check below this is also redundant, because the only way that this loop would complete is if the full path had been completed. There's no need for explicit validation. This `return` already ensures it.

src/main.ts
```diff
@@ -423,7 +426,20 @@ async function buildConvertPath (
         if (!format.to) continue;
         if (!format.mime) continue;
         if (path.some(c => c.format === format)) continue;
-        queue.push(path.concat({ format, handler }));
+        
+        const newPath = path.concat({ format, handler });
+        const pathSignature = newPath.map(c => c.format.format).join("→");
```
p2r3: Using `→` here is just ever so slightly unoptimal, as that's a Unicode character taking up more memory than an ASCII character. There are other ASCII characters that would be more fitting. `•` comes to mind, but there's probably others.

src/main.ts
mdev34-lab: Oh. Thanks for noting, I'll modify this right now!

src/handlers/meyda.ts
mdev34-lab: Done! Removed the redundant passthrough check. =)

src/main.ts
mdev34-lab: Done. Sorry for the nonsensical logic...

src/main.ts
mdev34-lab: Done. I changed the concatenation to the `•` character.

p2r3:
LGTM

Issue #75, opened 2026-02-16
Support more "Quite OK" formats


schrmh:
At least audio and a package archive format do exist as well:
https://github.com/phoboslab?tab=repositories&q=Quite+OK

mpopa4096-source:
This could work for QOA: https://www.npmjs.com/package/qoa-format

p2r3:
Closing as implemented thanks to the QOA PR. A separate issue can be made using the new guidelines for the archive format if you wish.

PR #216, opened 2026-02-17
Add Quite OK Audio (QOA) support

SladkyCitron:
Added Quite OK Audio (QOA) support based on [qoa-fu](https://github.com/pfusik/qoa-fu)
References #75


p2r3:
This suffers from the same problem as `qoi-fu`, that is, the TS transpilation uses a less strict config than we do. Thus, the project fails to build.

The way I fixed this for `qoi` was by including the JS version (not the TS one), and routing tsc away from the TS version. See "paths" in `tsconfig.json`.

SladkyCitron:
Made tsc use the JS transpilation instead of the TS one as you said, it should work now.

PR #219, opened 2026-02-17
exe bug fix

millennIumAMbiguity:
fixes for bat -> exe


src/handlers/batToExe.ts
```diff
@@ -1,5 +1,8 @@
 import type { FileData, FileFormat, FormatHandler } from "../FormatHandler.ts";
 
+import headUrl from "./batToExe/exe65824head.bin?url";
+import footUrl from "./batToExe/exe65824foot.bin?url";
```
p2r3, line 4: What's the `?url` query param for?

millennIumAMbiguity: It gets the URL from a file path. 
Removing the `?url` gives `Cannot find module`

millennIumAMbiguity:
fixes missed for #201

PR #225, opened 2026-02-17
ZIP to PNG (And other formats)

JulianHeuser:
I think one big component missing is conversion from ZIP to other formats. This opens a lot of pathways for other archive formats (like APK).

Currently, this is a simple implementation that allows for ZIP -> PNG by extracting and downloading only files that match the output format.

My plan is to run every file back through the converter recursively, so we essentially batch convert all files in the archive to the target format. I'm open to feedback and suggestions for alternatives though.

### DIFF
--- a/src/handlers/jszip.ts
+++ b/src/handlers/jszip.ts
@@ -1,5 +1,8 @@
 import type { FileData, FileFormat, FormatHandler } from "../FormatHandler.ts";
+import normalizeMimeType from "../normalizeMimeType.js";
+import mime from "mime";
 import JSZip from "jszip";
+import type { JSZipObject } from "jszip";
 
 class jszipHandler implements FormatHandler {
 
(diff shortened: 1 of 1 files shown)


p2r3:
Firstly, we already have a crude handler for ZIP files that renames them to other ZIP-like files. This seems to do the same, just with more steps in the middle.

Secondly, it doesn't make sense to approach this problem from the angle of "converting ZIP files". Archive formats are versatile by definition, there's no single handler that can meaningfully deal with them all.

Thirdly, if the goal was to provide an interface for extracting ZIP files, then that's something that should be dealt with system-wide. The current implementation doesn't support mixing file formats in routes, nor should it. ZIP files would break that convention.

All in all, I don't think this PR makes much sense. It seems to just be doing things for the sake of doing things.

Edit: upon rereading, it seems like the third point is the one that applies here. I.e., it seems like you're trying to create a solution for bundling input files. Again, that should be a system-wide change, not a handler. I'd even argue that such a change is likely unwelcome, because then you'd need to separate converting ZIP files and converting the contents of the ZIP files, which gets messy.

JulianHeuser:
I mean, this project gives the vibe of doing things for the sake of doing things. I get it though, I could see things getting messy if I went further with this. It was still useful to get a grasp of the code, I'll see if there's any other funny file formats I could add.

PR #228, opened 2026-02-17
Added .txt to .bat

millennIumAMbiguity:
converts text file to batch like this

```bat
@echo off
echo txt line 1
echo txt line 2
pause
```

Special characters in the batch are escaped, and empty lines are preserved.

### DIFF
--- a/src/CommonFormats.ts
+++ b/src/CommonFormats.ts
@@ -69,6 +69,13 @@ const CommonFormats = {
         "text/markdown",
         ["document", "text"]
     ),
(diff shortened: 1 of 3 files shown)


pixelkat5:
Seems good :)

millennIumAMbiguity:
The commit history got really scred up after rebasing, so I did a reset.

millennIumAMbiguity:
@p2r3 does it look good?
Based on your comment in #242, I think this PR may have been missed.
would be nice if it could be merged before I have to rebase it again

PR #230, opened 2026-02-17
Heuristic search based on Dijkstra's algorithm

Alexander1248:
Implementation of heuristic search based on Dijkstra's algorithm.

A weighted conversion graph (`TraversionGraph`) is introduced, where MIME types are represented as nodes and handler-based conversions as directed edges with computed costs. Pathfinding is performed using Dijkstra’s algorithm and returns conversion chains ordered by total accumulated cost.

Edge cost is calculated heuristically and takes into account:

* penalties for category changes (with configurable transitions and a default cost),
* an additional multiplier for lossy conversions,
* handler priority influence.

Supports a simple and advanced mode similar to old algorithm. Heuristic parameters are configurablе.

Graph construction from cache takes less than 1 second (measured ~25.40 ms for 468 nodes and 62 166 edges).

Added two new optional fields to `FileFormat`:

* **`category?: string | string[]`** — defines the logical category (e.g., image, video, audio) used for grouping formats and applying category-based penalties during pathfinding. Supports multiple categories per format.

* **`lossless?: boolean`** — indicates whether the format preserves data without quality loss.

The `lossless` flag acts as a *local criterion* for conversion evaluation: when a format is used as a **target (`to`) format**, this flag determines whether the conversion implies data loss. If `lossless` is `false` or undefined, the conversion is treated as lossy and penalized accordingly in the heuristic cost calculation.

### DIFF
--- a/src/TraversionGraph.ts
+++ b/src/TraversionGraph.ts
@@ -64,9 +64,11 @@ export class TraversionGraph {
                 toIndices.forEach(to => {
                     if (from.index === to.index) return; // No self-loops
                     let cost = DEPTH_COST; // Base cost for each conversion step
-                    if (from.format.category && to.format.category) {
-                        const fromCategories = Array.isArray(from.format.category) ? from.format.category : [from.format.category];
-                        const toCategories = Array.isArray(to.format.category) ? to.format.category : [to.format.category];
+                    const fromCategory = from.format.category || from.format.mime.split("/")[0];
+                    const toCategory = to.format.category || to.format.mime.split("/")[0];
+                    if (fromCategory && toCategory) {
+                        const fromCategories = Array.isArray(fromCategory) ? fromCategory : [fromCategory];
+                        const toCategories = Array.isArray(toCategory) ? toCategory : [toCategory];
                         if (CATEGORY_HARD_SEARCH) {
                             cost += CATEGORY_CHANGE_COSTS.reduce((totalCost, c) => {
                                 // If the category change defined in CATEGORY_CHANGE_COSTS matches the categories of the formats, add the specified cost. Otherwise, if the categories are the same, add no cost. If the categories differ but no specific cost is defined for that change, add a default cost.
@@ -83,8 +85,10 @@ export class TraversionGraph {
                             else cost += Math.min(...costs.map(c => c.cost)); // If multiple category changes are involved, use the lowest cost defined for those changes. This allows for more nuanced cost calculations when formats belong to multiple categories.
                         }
                     }
-                    else if (from.format.category || to.format.category) {
-                        cost += DEFAULT_CATEGORY_CHANGE_COST; // If one format has a category and the other doesn't, consider it a category change
+                    else if (fromCategory || toCategory) {
+                        // If one format has a category and the other doesn't, consider it a category change
+                        // Should theoretically never be encountered, unless the MIME type is misspecified
+                        cost += DEFAULT_CATEGORY_CHANGE_COST;
                     }
                     cost += PRIORITY_COST * formats.indexOf(from.format); // Add cost based on handler priority (lower index means higher priority)
                     if (!to.format.lossless) cost *= LOSSY_COST_MULTIPLIER; // If the output format is lossy or unspecified, apply the lossy cost multiplier


p2r3:
Ooh, _very_ exciting! I'll do some testing in just a bit.

p2r3:
Hm, didn't take long to run into an issue.

Simple, direct paths work fine. However, when I tried PDF to SVG, this algorithm declared that no path was found, while the existing BFS solution quickly found the PDF -> PNG -> SVG route.

I've only glanced at the code very briefly, so I can't reasonably guess at the cause.

p2r3:
Ah, actually, I do have _one_ guess - could it be that it's grouping nodes by _just_ MIME type? That wouldn't be valid, as we've got lots of different handlers doing different things with files.

The program first attempts PNG -> SVG through ImageMagick, which fails because IM is missing the required delegates. The _correct_ solution is routing the PNG -> SVG step through the svgTrace handler, but I assume that's been shadowed.

The existing BFS solution runs through _all_ of the valid handler input options, not just all of the MIME types.

p2r3:
Or... no, you do mention that the graph's edges are handler inputs/outputs, so that sounds good? It's just that I can't really see it trying anything other than ImageMagick in the logs...

Sorry for the ramble.

Alexander1248:
It tries all valid paths for all handlers.
I think category heuristics may be too strong, because they are greater than or equal to the depth cost (+1 per depth). I think the values for categories should be reduced by a factor of 5–10.

Alexander1248:
With the current costs, the search prefers to add depth rather than change categories, which can cause problems in some cases.

Alexander1248:
Also, I haven't marked the new handlers with the lossless and category parameters, and their costs are worse than the heuristics from yesterday's code.

p2r3:
> Also, I haven't marked the new handlers with the lossless and category parameters, and their heuristics are worse than the heuristics from yesterday's code.

Ah, this is probably the issue then. I was confused about why category differences would be an issue - both the failed route and the "correct" route are of the same depth and go through the same formats, so really the only difference is in which handler is preferred.

Yeah, I assume the svgTrace handler probably just has worse heuristics than ImageMagick.

p2r3:
Do you plan on going through the new handlers yourself and updating heuristics?

Alexander1248:
Yes, I am marking the new handlers and configuring the heuristics right now.

p2r3:
Hm, this is annoying.

Even with the new handlers marked, my previous PDF -> SVG test fails, and I don't think it's the fault of the heuristics. I think the problem is that certain handlers have inaccurate format lists. In this case, ImageMagick claims that it supports SVG output, when in reality it throws an error due to missing a delegate. Dijkstra's algorithm can't really take that into account, so it finds the "shortest" path through ImageMagick, but in practice that path doesn't actually exist.

This seems to imply that we'll have to manually comb through ImageMagick's and FFmpeg's supported formats and blacklist the ones that don't work in practice. Alternatively, we could weigh the order in which the handlers are imported more heavily, which approaches the way that the current system works.

p2r3:
I'm starting to realize that this has one big drawback to BFS. Yes, it's more optimal and faster in cases where it works, but it also introduces cases where it _doesn't_ work. That's pretty bad for a file converter. If we have the tools, there shouldn't be a scenario in which it just gives up and doesn't find the solution. Continuously tweaking the weights to make that not happen also sounds like a nightmare.

Maybe a solution like k-shortest paths would work better here? I.e., try the shortest path first, then the second shortest, then the third, and so on.

Files are strange and unpredictable - depending on the input, some files might not be compatible with certain handlers, despite that handler _technically_ supporting the input format. For example, some image and video formats have to be in specific resolutions. If Dijkstra's routed through those, then sometimes those paths would work, and sometimes they wouldn't. It's quite dangerous to not have a fallback.

Alexander1248:
Dijkstra already tries all paths in order of cost. I also found an error. For some reason, it tries to convert PDFs using `ImageMagick`, but it doesn't try to traverse through `pdftoimg`, even though `pdftoimg` works for PDF → PNG conversions.

Alexander1248:
False alarm. It actually uses the path `pdftoimg(application/pdf) → pdftoimg(image/jpeg) → svgTrace(image/svg+xml)` but fails in pdf → jpeg conversion:
```
pdftoimg pdf → jpeg Invalid output format.  
TraversionGraph.ts:131 Found path at iteration 534 with cost 3.71: pdftoimg(application/pdf) -> pdftoimg(image/jpeg) -> svgTrace(image/svg+xml)  
TraversionGraph.ts:133 Path valid! Yielding path: application/pdf → image/jpeg → image/svg+xml  
```
Bug somewhere in pdftoimg pdf → jpeg also in my search at the same time. 
In the master branch, the PDF → JPEG route uses PDF → PNG → JPEG.

p2r3:
> Dijkstra already tries all paths in order of cost

Does it? My understanding was that it finds just the shortest path to each node from the start, leading to it eliminating "suboptimal" paths. Again, I haven't really looked into how you've implemented it.

> Bug somewhere in pdftoimg pdf → jpeg.

Huh? How does the current implementation succeed then? I mean, sure, the route that BFS uses is PDF -> PNG which, shouldn't Dijkstra also be using that because it's less lossy? And, again, if it really does consider all paths in order of weight, then it _should_ be attempting PDF -> PNG as well, not just JPEG.

p2r3:
And this just kinda further pushes the point I was talking about earlier - if there's a problem in one edge, like there is in PDF - JPEG right now, then Dijkstra's will just assume that route is impossible, because the alternative would've been longer, so it didn't even consider it. It'll be hell if we have to chase every single inconsistency for the converter to work.

Alexander1248:
I implemented Dijkstra using an AsyncGenerator so that it yields all possible paths in order of cost.

p2r3:
> yields all possible paths in the graph in order of cost.

Then there shouldn't be a case where BFS succeeds and your algorithm doesn't, right? If it considers every path, then it should have the same success rate as BFS, with the benefit of being heuristically guided. But evidently, that's not the case. I've re-ran my PDF → SVG test three times now on your fork, and it claims there's no valid route. The current BFS implementation succeeds, so clearly a route _does_ exist, you're just eliminating it somehow.

p2r3:
For the record, the individual steps (PDF -> PNG, then PNG -> SVG) work just fine. It's the combination that doesn't, because it's never attempted.

Alexander1248:
I think I know why this happens. I filter visited nodes only by their index, so it tries only one handler and blocks the node.

Alexander1248:
Now it checks only the visited nodes in the state they were in when the node was added to the queue. This causes blocking of visits only for subsequent nodes, not the current one, and at the same time does not lead to an explosion of options, as would happen if filtering were done by index and handlers.

p2r3:
Back to exciting again! After running a few more tests now, it does seem like the issue is fixed, and in general the routing is much more sensible than BFS.

Though, one last thing - do you think a manual bias would be a good idea? For example, converting DOCX to PNG currently spits out an image with text from the input document rendered on one long line. It _should_ return a screenshot of the document instead.

From the algorithm's point of view, this makes perfect sense. The route (docx - text - png) is short and only changes category once, whereas the "correct" route (docx - html - svg - png) is far more convoluted. Would a manual bias be a good idea here?

Alexander1248:
I think so. Also, this might be solved by assigning the correct categories. For example: `docx` is `document`; `html` is `document` and `text`; `svg` is `image`, `vector`, and `text`; and `png` is `image`. This way, transitions will be smooth, because there are no category changes (if `CATEGORY_HARD_SEARCH` is disabled).

p2r3:
Hm, good point.

Are you planning on implementing anything else right now, or am I good to do a code review, some final touchups, and then merge?

Alexander1248:
`CATEGORY_HARD_SEARCH` uses differential weighting, which is the sum of all differing categories. It is more strict and is disabled by default in the search.

Alexander1248:
I’m not planning to do anything yet, just thinking about making the heuristic parameters editable externally.

p2r3:
> I’m not planning to do anything yet, just thinking about making the heuristic parameters editable externally.

If you do come around to that, feel free to open a new PR. I'll be merging this one as-is to move on to other pull requests, as otherwise everything more or less depends on this.

Issue #160, opened 2026-02-17
Support for .cur (windows cursor) files to other image formats


NeoByt3:
The .cur file format is almost identical to .ico and can easily be added to the website. I could not find any repositories or open source libraries allowing .cur files to be converted to other file types.
"The CUR file format is an almost identical image file format for non-animated [cursors](https://en.wikipedia.org/wiki/Cursor_(computing)) in Microsoft Windows. The only differences between these two file formats are the bytes used to identify them and the addition of a hotspot in the CUR format header"[Taken from here]( https://en.wikipedia.org/wiki/ICO_(file_format))

pompom454:
I actually just [made a repo](https://github.com/pompom454/cur-png.js) for this, specifically JUST to satisfy this issue.

pixelkat5:
This should be quick, already working on it

ConnorTippets:
implemented in #377

PR #233, opened 2026-02-18
Added support for Cur files, fulfilling Issue 160

pixelkat5:
[Link to the Issue](https://github.com/p2r3/convert/issues/160)

**Small overview of what I did:**
added `	case "image/x-win-cursor": return "image/x-win-cursor";` to normalizeMimeType.ts inside src/normalizeMimeType.ts
added Curhandler.ts to scr/handers/

### DIFF
--- a/src/handlers/CurHandler.ts
+++ b/src/handlers/CurHandler.ts
@@ -1,0 +1,232 @@
+import type { FormatHandler, FileFormat, FileData } from "../FormatHandler.js";
+
+// ── internal helpers ───────────────────────────────
+
+async function curToPng(bytes: Uint8Array): Promise<Uint8Array> {
+  const buf = bytes.buffer.slice(bytes.byteOffset, bytes.byteOffset + bytes.byteLength);
+  const dv = new DataView(buf);
+
+  const type = dv.getUint16(2, true);
+  if (type !== 2) throw new Error("Not a CUR file");
+
+  const imageSize   = dv.getUint32(14, true);
+  const imageOffset = dv.getUint32(18, true);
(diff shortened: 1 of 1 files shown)


pixelkat5:
The css update was for something else by the way.

src/handlers/index.ts
```diff
@@ -26,6 +26,7 @@ import flptojsonHandler from "./flptojson.ts";
 import floHandler from "./flo.ts";
 import batToExeHandler from "./batToExe.ts";
 import textEncodingHandler from "./textEncoding.ts";
+import CurHandler from "./CurHandler.ts";
```
p2r3, line 29: Handler imports/exports should be in camelCase. That is, first letter should be lowercase.

src/handlers/CurHandler.ts
```diff
@@ -0,0 +1,232 @@
...
+      internal:  "jpg",
+    },
+    {
+      name:      "WebP (CurHandler)",
+      format:    "webp",
+      extension: "webp",
+      mime:      "image/webp",
+      from:      true,
+      to:        true,
+      internal:  "webp",
```
p2r3, line 188: Consider syncing with upstream and replacing these with `CommonFormats` equivalents. If you don't know what I mean, check out the format list of `canvasToBlob.ts` on the main branch, for example.

pompom454: I would also like to add that BMP data is possible inside a .CUR and has different handling. If you don't know how to implement this, check my cur-png.js repo.

p2r3:
Superseded by #377

PR #236, opened 2026-02-18
Added CgBI to PNG

amythiz:
Added support for converting from [PNG CgBI Format](https://theapplewiki.com/wiki/PNG_CgBI_Format), non-standart PNG format used in Apple devices (mostly in early iOS apps but it is still supported on modern Apple devices), to normal PNG. Based on [cgbi-to-png](https://github.com/jakubknejzlik/cgbi-to-png)


amythiz:
Tested with these files:
[test-cgbi.zip](https://github.com/user-attachments/files/25376833/test-cgbi.zip)

src/handlers/index.ts
```diff
@@ -24,6 +24,7 @@ import nbtHandler from "./nbt.ts";
 import peToZipHandler from "./petozip.ts";
 import flptojsonHandler from "./flptojson.ts";
 import floHandler from "./flo.ts";
+import CgbiToPngHandler from "./cgbi-to-png.ts";
```
p2r3: Handler imports/exports should be in camelCase. The first letter should be lowercase.

src/handlers/cgbi-to-png.ts
```diff
@@ -0,0 +1,206 @@
...
+    },
+    {
+      name: 'Portable Network Graphics',
+      format: 'png',
+      extension: 'png',
+      mime: 'image/png',
+      from: false,
+      to: true,
+      internal: 'png'
+    }
```
p2r3: Consider syncing with upstream and replacing this with its CommonFormat equivalent. See `canvasToBlob.ts` on the main branch for an example. Also, `cgbi-png` would likely need `lossless: true` specified.

We've changed quite a bit since this PR was introduced 😅

PR #240, opened 2026-02-18
added .sb3 to .html converter

WFlyToTheSky:
i added a scratch sb3 file to png  converter

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,4 +4,3 @@ node_modules
 start.sh
 dist
 cache.json
-.idea


mmewtwosaysbye:
why to png not exe

WFlyToTheSky:
> why to png not exe

i tried to do to html or something but failed, and so i decided to just go with this for now, but i might make one that converts to that one day

p2r3:
There's a bunch of garbo here that is most definitely out of place. You've committed config files generated by your IDE, and you've added a bunch of dependencies ~~as developer dependencies~~? Strange ones too - ones that I wouldn't expect to see on a browser-based project.

WFlyToTheSky:
I'm pretty new to this stuff, what developer dependencies are you talking about?

p2r3:
- You've commited the `.idea` directory, presumably generated by your IDE. That shouldn't be there.
- You've included `@types/node` under dev dependencies. This project is developed and built with Bun, so Node TS support shouldn't be necessary.
- You also include `canvas` and `html2canvas`, which are redundant. Not only is this already a browser environment, meaning we do not need a package that emulates a canvas, but this is also a dynamic file converter. You can just output the HTML, and if the user wanted a PNG, our other handlers will be able to convert that HTML to a PNG. You don't need to handle this logic yourself with heavy dependencies.
- You also include a submodule for `turbowarp-packager` _and_ an npm dependency for seemingly the same thing? I'm not familiar with these libraries, but this looks strange.
- I'm also not entirely sure if the changes in `vite.config.js` are necessary - I don't see why the bundler couldn't do that on its own - but again, I'm not familiar with these libraries.
- As for the code, I've only glanced over it, but I do have a few nitpicks there too. You write your own MIME type resolver functions, when we already include the `mime` library for this purpose. You also don't use `normalizeMimeType` as the README suggests you should. You also use the `meyda` handler in there to convert sound to images, which... why?? That's incredibly arbitrary? Again, just generate a normal HTML file with `<audio>` tags embedded as base64 or something. That's both simpler and more modular.

p2r3:
Also, forgot to mention, your fork fails to build. There are type errors. (You can test these with `bun run build` on Bun, or `npm run build` on Node, though as I said, you _should_ ideally be using Bun for this.)

WFlyToTheSky:
i added turbowarp packager stuff from when i was trying to originally compile to an html and forgot to remove it

WFlyToTheSky:
i'm pretty sure i fixed everything, and it successfully builds

PR #241, opened 2026-02-18
Add ZSTD support using fzstd

chippytech:
## Add Zstandard (.zst) support via fzstd

This PR adds support for Zstandard compression and decompression using the lightweight `fzstd` package.

### What’s included

* New `zstdHandler`
* Supports:

  * Compress: any input → `.zst`
  * Decompress: `.zst` → original bytes
* Fully client-side
* No mutation of input buffers
* Matches existing handler architecture (modeled after `jszipHandler`)

### Why

Zstandard offers:

* Better compression ratios than gzip/zip
* Faster compression/decompression
* Growing ecosystem adoption

This makes the converter more flexible for modern compression workflows while keeping everything browser-based.

### Implementation Notes

* Uses `fzstd` (WASM-backed, runs in browser)
* Operates directly on `Uint8Array` (compatible with existing `FileData`)
* No changes to core architecture
* No impact on existing handlers

Happy to adjust!


src/handlers/zstdHandler.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  public supportedFormats: FileFormat[] = [
+    {
+      name: "Zstandard Compressed File",
+      format: "zst",
+      extension: "zst",
+      mime: "application/zstd",
+      from: true,
+      to: true,
+      internal: "zst"
```
p2r3, line 16: Consider syncing with upstream and adding `lossless: true` as required since #230. Well, not "required", but definitely advisable given that this is a lossless conversion.

p2r3:
Your PR fails to build:
> `src/handlers/zstdHandler.ts(2,10): error TS2724: '"fzstd"' has no exported member named 'compress'. Did you mean 'decompress'?`

PR #242, opened 2026-02-18
Text to BAT and SH scripts

JulianHeuser:
Adds converters for text to SH (Linux script) and text to BAT (Windows script)

These are fairly straightforward, they just print the content. The BAT converter is more complex since it has to escape special characters.

The fun part is the connections this opens up. For example:
image

And the BAT (testing through wine):
image

### DIFF
--- a/src/handlers/textToBat.ts
+++ b/src/handlers/textToBat.ts
@@ -1,0 +1,66 @@
+import type { FileData, FileFormat, FormatHandler } from "../FormatHandler.ts";
+
+class textToBatHandler implements FormatHandler {
+
+  public name: string = "textToBat";
+
+  public supportedFormats: FileFormat[] = [
+    {
+      name: "Plain Text",
+      format: "text",
+      extension: "txt",
+      mime: "text/plain",
+      from: true,
+      to: false,
+      internal: "text"
+    },
+    {
+      name: "Windows Batch file",
+      format: "batch",
+      extension: "text/bat",
+      mime: "text/windows-batch",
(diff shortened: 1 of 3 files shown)


src/handlers/textToBat.ts
```diff
@@ -0,0 +1,66 @@
...
+    },
+    {
+      name: "Windows Batch file",
+      format: "batch",
+      extension: "text/bat",
+      mime: "text/windows-batch",
+      from: false,
+      to: true,
+      internal: "bat"
+    },
```
p2r3, line 25: Consider syncing with upstream and replacing plaintext with the `CommonFormats` equivalent, as well as flagging these as `lossless: true` (introduced in #230). If you don't know what I mean, check out the format list of `canvasToBlob.ts` on the main branch, for example.

ConnorTippets:
Mentioning #228 as it's related (txt -> bat)

millennIumAMbiguity:
I suggest that you remove your own .txt -> .bat in favor of #228
Making the pr a pure .txt -> .sh

p2r3:
Yeah, sorry, I decided to merge #228 for BAT instead, as that PR was more up-to-date with the new format labeling structure. Feel free to reopen a new PR for just .sh if you wish.

JulianHeuser:
Made the SH pr: https://github.com/p2r3/convert/pull/296

Also I may be missing something but it looks like the batch handler wasn't added to the main file? It doesn't appear in the output options either.

Issue #162, opened 2026-02-17
Support for .wld to .png


orrinpants:
Somebody else made this request, but they didn't say how or what to. 
Convert .wld (terraria world) to .png
[TerraMap source code](https://github.com/TerraMap/terramap.github.io) this is source for a tool that parses .wld as an image

ConnorTippets:
~~Wrong repo. That's just a website.~~
~~You're likely thinking of [TerraMap/windows](https://github.com/TerraMap/windows).~~

Nevermind, that repo does contain a js-based .wld parser!

pixelkat5:
I could try working on something like this :)

pixelkat5:
> I could try working on something like this :)

I will say that I _am_ working on another feature as of now so this may take a bit due to it being postponed.
Though I do think this shouldn't be very hard.

pixelkat5:
Almost done :D

pixelkat5:
<img width="8400" height="2400" alt="Image" src="https://github.com/user-attachments/assets/8d26548e-e549-4601-8447-b7b424d916d7" />
<img width="1477" height="910" alt="Image" src="https://github.com/user-attachments/assets/f08653be-16ef-4a8e-bd1e-b163d2a3d5fc" />

Lets go I got it working!

pixelkat5:
I think this can now be closed.

p2r3:
> I think this can now be closed.

@pixelkat5 No, I've requested changes on both of your open pull requests.

PR #244, opened 2026-02-18
Terraria wld support

pixelkat5:
im hoping thisll work this time?

### DIFF
--- a/src/handlers/Wldhandler.ts
+++ b/src/handlers/Wldhandler.ts
@@ -1,0 +1,1488 @@
+import type { FormatHandler, FileFormat, FileData } from "../FormatHandler.js";
+
+// ── Base file parser ──────────────────────────────────────────────────────────
+
+class TerrariaFileParser {
+  offset: number = 0;
+  buffer!: DataView;
+  options: { ignoreBounds: boolean } = { ignoreBounds: false };
+  RLE: number = 0;
+
+  async loadFile(file: File): Promise<void> {
+    const buffer = await new Promise<ArrayBuffer>((resolve, reject) => {
+      const reader = new FileReader();
+      reader.onload  = () => resolve(reader.result as ArrayBuffer);
+      reader.onerror = () => { reader.abort(); reject(reader.error); };
+      reader.readAsArrayBuffer(file);
+    });
+    this.buffer = new DataView(buffer);
+  }
+
+  readUInt8(): number {
+    this.offset += 1;
+    if (this.options.ignoreBounds && this.offset > this.buffer.byteLength) return 0;
+    return this.buffer.getUint8(this.offset - 1);
+  }
+  readInt16(): number {
+    this.offset += 2;
+    if (this.options.ignoreBounds && this.offset > this.buffer.byteLength) return 0;
+    return this.buffer.getInt16(this.offset - 2, true);
+  }
+  readUInt16(): number {
+    this.offset += 2;
+    if (this.options.ignoreBounds && this.offset > this.buffer.byteLength) return 0;
+    return this.buffer.getUint16(this.offset - 2, true);
+  }
+  readInt32(): number {
+    this.offset += 4;
+    if (this.options.ignoreBounds && this.offset > this.buffer.byteLength) return 0;
+    return this.buffer.getInt32(this.offset - 4, true);
+  }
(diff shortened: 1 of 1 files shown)


mmewtwosaysbye:
what do css styles have to do with terraria?

src/handlers/Wldhandler.ts
```diff
@@ -0,0 +1,1488 @@
...
+
+  doConvert: async (
+    inputFiles: FileData[],
+    _inputFormat: FileFormat,
+    _outputFormat: FileFormat,
+  ): Promise<FileData[]> => {
+    const results: FileData[] = [];
+
+    for (const file of inputFiles) {
+      window.showPopup("<h2>Parsing world file...</h2><p>This may take a moment for large worlds.</p>");
```
p2r3, line 1415: No, no no! Do not create popups from within handlers!!! This is backend code, it should _not_ touch the frontend _ever_!

p2r3:
Where's this code from, by the way?

ConnorTippets:
> Where's this code from, by the way?

TerrariaWorldParser in [handlers/WldHandler.ts](https://github.com/pixelkat5/convert-cur-update/blob/4789d5f9aad8165f9f06189464ef853c715dfc6f/src/handlers/Wldhandler.ts#L135) seems to match almost 1:1 with terrariaWorldParser from [cokolele/terraria-world-file-js/src/node/terraria-world-parser.js](https://github.com/cokolele/terraria-world-file-js/blob/d719ef8a152a5c8c83c6977a63c09f0791891512/src/node/terraria-world-parser.js#L4)

p2r3:
Then it should be included as a submodule, not ripped without proper licensing or attribution.

ConnorTippets:
> Then it should be included as a submodule, not ripped without proper licensing or attribution.

To be fair, that repo is a public archive, so it could be pulled at any moment. Plus, it's under MIT. Attribution would be nice though.

ConnorTippets:
Also, #162 should probably be mentioned here.

src/normalizeMimeType.ts
```diff
@@ -11,6 +11,7 @@ function normalizeMimeType (mime: string) {
     case "audio/x-quicktime": return "video/quicktime";
     case "audio/x-flo": return "audio/flo";
     case "application/x-flo": return "audio/flo";
+    case "application/x-terraria-world": return "application/x-terraria-world";
```
ConnorTippets, line 14: Is this necessary? To my eye, it seems to be outputting what is inputted, which the function already does.

p2r3:
> To be fair, that repo is a public archive, so it could be pulled at any moment.

Any repo can be pulled at any moment. That doesn't mean you should rip their code.

> Plus, it's under MIT.

I don't think that's an issue? MIT code can be included in GPL code AFAIK.

b1ek:
> That doesn't mean you should rip their code.

id rather say its important to include it as a module since it will make a horrible mess out of the codebase if you always do that instead of `yarn add X`

p2r3:
Marking this as draft until the issues get resolved.

Celestemysterios:
hey i do wanna say i found this thing that supports more modern versions of terraria/seems to be updated more frequently
https://github.com/TEdit/terraria-world-file-ts
it is in typescript though not js (idk what typescript is)
i would figure out how to implement this myself but idk js/ts and how file conversation work in general so uhh🥲

ConnorTippets:
@Celestemysterios This project uses typescript. So yes, we could use this.

ConnorTippets:
this uhh... hasn't moved for a while.

Celestemysterios:
bit scared
i wish i could help because j really want this to be in the thing but again, idk JavaScript or typescript

ConnorTippets:
i would willingly implement this in a heartbeat, but i don't wanna step over this person's work

Celestemysterios:
thats fair
hopefully they come back and implement it thenselves
how much longer do you plan on waiting?

ConnorTippets:
@pixelkat5 you making a move on this anytime soon? we got some issues for you to fix...

src/handlers/Wldhandler.ts
lyqht: Nonono

PR #246, opened 2026-02-18
Format definition categories and flags, more common formats

BoBkiNN:
This PR continues #206 and adds several updates:
* Added Category object for common categories
* Added override option for `supported(...)` method
* Added back comments explaining used lossless flag
* Builder with method chaining in addition to positional parameters
* Added more common format definitions

TODO:
* Add more document common formats

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,11 +99,11 @@ class dummyHandler implements FormatHandler {
 
   async init () {
     this.supportedFormats = [
-      // Create a FileFormat based on the common PNG definition.
-      // "png" is this handler's internal format identifier.
-      // Both `from` and `to` are set to false, meaning this handler
-      // does not support conversion from or to PNG.
-      CommonFormats.PNG.supported("png", false, false)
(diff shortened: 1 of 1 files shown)


p2r3:
I get that this is WIP, I've only got one small nitpick so far - I don't think there's a need to over specify if a format is lossy.

If it's lossless, add the `true`. If it's lossy, just leave the argument blank. Makes it a bit easier to tell at a glance what is or isn't lossless.

BoBkiNN:
xlsx format has different categories in different handlers, is this intended?

p2r3:
> xlsx format has different categories in different handlers, is this intended?

I think all of the Office-esque files should have the "document" category. Not entirely sure why that's not the case everywhere.

Oh, sorry, misunderstood. You were saying that XLSX is sometimes _missing_ the "spreadsheet" category. This isn't egregious, but yes, it should just be the same everywhere.

PR #260, opened 2026-02-18
feat(frontend): progress on tool loading

ls-root:
When initalizing the handlers it shows a progress bar and which handler is currently loading.

Video:

https://github.com/user-attachments/assets/b1dd3d1e-31c7-4de1-8d91-564afd936719

### DIFF
--- a/index.html
+++ b/index.html
@@ -43,7 +43,8 @@
 
   <div id="popup-bg"></div>
   <div id="popup">
-    <h2>Loading tools...</h2>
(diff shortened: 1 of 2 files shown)


p2r3:
This doesn't make much sense to push to production. In production, we have a format list cache, so handlers are only initialized while routing.

Correct me if I'm wrong, but I believe this feature only applies when you haven't set up `cache.json` (as mentioned in the readme).

PR #261, opened 2026-02-18
Add support for tracker audio files like ProTracker and Scream Tracker

iAmInActions:
I have started work on adding libopenmpt to allow for rendering sample+pattern based audio formats into raw audio which can then be converted into other formats. This mostly already works, the import side is 100% functional for all formats supported by libopenmpt but the out output still has issues:

Formats added by this handler (input only for obvious reasons):
```
mptm, mod, s3m, xm, it, 667, 669, amf, ams, c67, cba, dbm, digi, dmf, dsm, dsym, dtm, etx, far, fc, fc13, fc14, fmt, fst, ftm, imf, ims, ice, j2b, m15, mdl, med, mms, mt2, mtm, mus, nst, okt, plm, psm, pt36, ptm, puma, rtm, sfx, sfx2, smod, st26, stk, stm, stx, stp, symmod, tcb, gmc, gtk, gt2, ult, unic, wow, xmf, gdm, mo3, oxm, umx, xpk, ppm, mmcmp
```

- Despite tweaking FFMPEGs cost routes to prioritize WAV, it still chooses to go over PNG in some cases if the output isnt a WAV file, this really shouldnt happen but im unsure why audio in -> audio out always runs into audio in -> audio to image -> image to video -> video to audio -> audio out.
- The handler is currently added as a file inside the repository, there don't seem to be official wasm repositories so unless it is to be built at project compile time, I don't see a better way of including it.

I hope this is fine for now, otherwise I might have to add a script for cloning and building the source tree during the compose process.


p2r3:
> Despite tweaking FFMPEGs cost routes to prioritize WAV, it still chooses to go over PNG in some cases if the output isnt a WAV file

That _is_ strange. I know we had an issue with this earlier, but it should be resolved by now. The cost for going from one audio format to another should be basically free compared to hopping mediums. Can you describe the exact scenario? Maybe in a way that I can test?

iAmInActions:
(also apologies for the slightly cursed inline patch in the libopenmpt minified js, i just added two calls so it works with the existing structure)

iAmInActions:
> > Despite tweaking FFMPEGs cost routes to prioritize WAV, it still chooses to go over PNG in some cases if the output isnt a WAV file
> 
> That _is_ strange. I know we had an issue with this earlier, but it should be resolved by now. The cost for going from one audio format to another should be basically free compared to hopping mediums. Can you describe the exact scenario? Maybe in a way that I can test?

I was testing this with aryx.s3m earlier, selected scream tracker input and mp3 output, it gave me a path of s3m -> wav -> gif -> mp4 -> webm -> mp3

Edit:
I forgot to mention my build process, I used docker compose onto a ramdisk as I did not want node.js installed on my pc

p2r3:
That should mean that simply converting MP3 to WAV would do the same thing, but that's not the case. It does route straight to WAV through FFmpeg.

I'll clone your fork in a moment to test, can you give me an example input file?

p2r3:
^ Never mind that, got a download for aryx.s3m

iAmInActions:
> That should mean that simply converting MP3 to WAV would do the same thing, but that's not the case. It does route straight to WAV through FFmpeg.
> 
> I'll clone your fork in a moment to test, can you give me an example input file?

yeah, here you go:
https://modarchive.org/index.php?request=view_by_moduleid&query=191789

iAmInActions:
Once this is merged I wanna start working on a midi to wav converter (and possibly also wav to midi using the existing wav to png as a base and maybe also include some other more obscure audio **output** options from the https://buzzer-studio.atomic14.com/ codebase)

p2r3:
Uh no, sorry, I can't reproduce this. On your fork, it routes directly s3m - wav - mp2 (should be mp3 but that's a different can of worms).

p2r3:
Could it be that FFmpeg is somehow busted on your setup?

iAmInActions:
> Could it be that FFmpeg is somehow busted on your setup?

Possible, I tested both in Opera and Firefox so I don't think the chromium engine is to blame for this which leaves only my docker version as a possible cause. Ill try rebuild it in a clean fedora distrobox and see if it still happens.

iAmInActions:
Ok, tested on a clean fedora with docker and with node under void, both worked without my issue being reproducible.
So I'm guessing docker is just broken in some weird way on void that makes ffmpeg specifically not work.

p2r3:
Great. In that case, this looks good to me.

PR #263, opened 2026-02-18
Renamed "jpg" format to "jpeg" in ImageMagic handler

amythiz:
Noticed that there are both JPEG and JPG formats avaliable in simple mode. Turns out jpeg is called "jpg" in ImageMagic, while it is called "jpeg" everywhere else. This PR should fix this problem. Attached screenshots of how it was before.

Снимок экрана 2026-02-18 164033

Снимок экрана 2026-02-18 164045


amythiz:
Now there's only one jpeg

Снимок экрана 2026-02-18 174025

PR #266, opened 2026-02-18
Editable categories for search

Alexander1248:
* Categories change costs are now editable. Handlers can now add custom, case-specific categories.
* Minor cost reconfiguration to compensate for the lossless multiplier. This was an attempt to fix p2r3/convert/issues/247, but it didn’t resolve the issue.
* Experimental path routing listeners (I tried to visualize the conversion graph at runtime, but Mermaid is too slow).


src/TraversionGraph.ts
```diff
@@ -142,32 +204,41 @@ export class TraversionGraph {
...
                 let edge = this.edges[edgeIndex];
                 const indexInVisited = visited.indexOf(edge.to.index);
                 if (indexInVisited >= 0 && indexInVisited < current.visitedBorder) return;
                 const handler = handlers.find(h => h.name === edge.handler);
                 if (!handler) return; // If the handler for this edge is not found, skip it
+                
                 queue.add({
                     index: edge.to.index,
-                    cost: current.cost + edge.cost,
+                    cost: current.cost + (edge.to.index === toIndex ? 0 : edge.cost),
```
Alexander1248: I'm not entirely sure about this change. It helps the traversal run faster, but it ignores suboptimal solutions at the last node before the destination.

p2r3:
We should really get some unit tests... Merging PRs like these is getting a bit scary, the last one managed to break a bunch of paths that had to be slowly rebalanced. I don't doubt that this will also mess something up ever so slightly.

I don't think I fully get the handler-specific categories thing - can you give an example of where that would be used?

Alexander1248:
I think unit tests would be especially helpful in this case. Currently, new handlers may introduce chaotic conversion variants that cause data loss. I think unit tests could partially mitigate this problem.

Handler-specific categories are more of a solution for the Meyda image <-> audio case. We can assign a handler-specific category cost to the Meyda handler, which can be used to avoid selecting Meyda as a transitional conversion when it’s not necessary.

p2r3:
That's cool, but we're actually currently struggling with the opposite problem as well.

There's an ongoing issue right now where Meyda isn't being picked for something like PNG -> MP3. The route requires doing (png - wav - mp3), and that first step is considered super lossy because it goes from an image to audio. Instead, it tries to route through FFmpeg, because (image - video - audio) _looks_ like a better deal, but it's not. It can't work, because FFmpeg will always error out when converting a video with no audio channel to an audio file.


So basically, the problem is that (image - video) and (video - audio) are normally good individually, but they're _very_ bad in combination. I don't see a way that handler-specific categories can reasonably fix this, as this problem requires context past any two nodes. Do you have any other ideas for solving this cleanly?

Alexander1248:
I’m currently fixing bugs in handler-specific categories — they don’t work as intended.

 I think the image → video → audio path issue can be solved by using a more adaptive cost function for unintended routes and better cost tuning. My current cost configuration in test environment solves this problem:

> Found path at iteration 73367 with cost 5.46:
> svgTrace(image/png) → meyda(audio/wav) → FFmpeg(audio/mpeg)

Parameters:
```
categoryChangeCosts: [
        {from: "image", to: "video", cost: 0.2}, // Almost lossless
        {from: "video", to: "image", cost: 0.4}, // Potentially lossy and more complex
        {from: "image", to: "audio", cost: 1.4}, // Extremely lossy
        {from: "image", to: "audio", handler: "ffmpeg", cost: 100}, // FFMpeg can't convert images to audio
        {from: "audio", to: "image", handler: "ffmpeg", cost: 100}, // FFMpeg can't convert audio to images
        {from: "audio", to: "image", cost: 1}, // Very lossy
        {from: "video", to: "audio", cost: 1.4}, // Might be lossy 
        {from: "audio", to: "video", cost: 1}, // Might be lossy
        {from: "text", to: "image", cost: 0.5}, // Depends on the content and method, but can be relatively efficient for simple images
        {from: "image", to: "text", cost: 0.5}, // Depends on the content and method, but can be relatively efficient for simple images
]

DEPTH_COST: 1
DEFAULT_CATEGORY_CHANGE_COST: 0.6
LOSSY_COST_MULTIPLIER: 1.4
HANDLER_PRIORITY_COST: 0.2
FORMAT_PRIORITY_COST: 0.05
```

p2r3:
I'm not gonna fiddle with the costs too much for now in fear of breaking something else, I'll just push a dirty patch that solves this very explicitly. Good luck fixing the handler-specific categories.

Alexander1248:
One possible solution is to introduce adaptive costs in the search that detect predefined sequences of category transitions and apply a penalty to those patterns.

For example, if the path follows `a → b → c` or `a → b → b → c`, an additional `+x` cost can be applied based on the pattern `[a, b, c]`.

p2r3:
Yeah, I did something similar in f4936d556efdc39a715dc0f7859bb0b74f7ce147 just now, but instead of at the graph level, it's done at the path attempt level.

It's a crude patch, and it introduces merge conflicts (sorry!), but people were complaining about one of the core mechanics of the tool not working so I had to put something out for now.

Alexander1248:
Okay, no problem! Then I’ll generalize this approach

src/TraversionGraph.ts
```diff
@@ -1,3 +1,4 @@
+import { int } from "three/tsl";
```
p2r3: Stray import?

Alexander1248: yes

Alexander1248:
Which testing framework will be used in this repository? Vitest?

p2r3:
I was just going to use Bun's built-in test runner after the deployment build has been created.

Alexander1248:
Ok, then I’ll write the tests using Bun

p2r3:
Oh, were you planning on writing the tests? I had already self-assigned #285 😅

I'm perfectly fine with letting you do it if you wish, I can find something else to work on. I've only just started.

Alexander1248:
I’m currently adding basic tests for `TraversingGraph` along with some mocks. Covering anything beyond the core search logic is hard at the moment because separating the backend from the window layer would require significant refactoring, or mocking the window object (I haven’t tried this and I’m not sure if it’s possible).

p2r3:
Ah, okay, so you're writing tests for the graph specifically? That makes sense. I'll continue writing general-purpose tests for actual conversion success with puppeteer then, I don't think it hurts to have both.

Alexander1248:
Ok, I forgot about Puppeteer 😅

Alexander1248:
I think this is the last change for this PR, apart from future fixes

PR #286, opened 2026-02-18
chore(tsconfig): enable unused locals and parameters checks

thesenegalesehitch:


thesenegalesehitch:
Enable noUnusedLocals and noUnusedParameters in tsconfig.json and fix resulting errors

PR #296, opened 2026-02-18
Text to SH script

JulianHeuser:
Adds a converter from text to a shell script, similar to the text to batch converter.

image


p2r3:
I also added the missing `bat->txt` handler as part of this PR. Not exactly super structurally clean to do it here, but it does the job.

PR #297, opened 2026-02-18
updated README.md to include an example of a custom defined format.

WFlyToTheSky:
since the new readme only included an example on how to define a format using the already defined commonformats, i updated it to show how to do it for other formats.


p2r3:
The indentation is here is messy. I'll just do this myself on the main branch, but thanks for pointing it out.

PR #299, opened 2026-02-18
I made a txt to infinite craft save file converter

WFlyToTheSky:
it basically just gets all the different words and turns them each into words in infinite craft, and gives them a random emoji


WFlyToTheSky:
I wasn't really sure what the mime type should be, so I just went with application/x-infinite-craft-ic

.gitignore
```diff
@@ -4,3 +4,4 @@ node_modules
 start.sh
 dist
 cache.json
+.idea
```
ConnorTippets: Why?

WFlyToTheSky: What do you mean?

WFlyToTheSky: I added it so the files from my IDE wouldn't show up, should I not do that?

ConnorTippets: Ideally, no. Your IDE is unique to you, and most likely won't be much use in the repo gitignore. Your personal install of git on your machine has its own global gitignore, which is preferred in most FOSS cases.
Create a file somewhere in your user profile, and then use the command `git config --global core.excludesfile <path>` where `<path>` is the path to the file you created.

WFlyToTheSky: I think I fixed it

SPLATPLAYS:
Just want to note, the save files for this applications are json

WFlyToTheSky:
They're zipped json files

p2r3:
This should probably have a more descriptive name. "txttoic" does not immediately suggest Infinite Craft, and for someone unfamiliar with the game, it might be hard to judge what the handler is for at a glance.

WFlyToTheSky:
I fixed it, it's now txtToInfiniteCraft

Issue #180, opened 2026-02-17
Text to audio conversion not implemented


copokasasagi:
<img width="280" height="140" alt="Image" src="https://github.com/user-attachments/assets/8656ad3a-047f-4243-b00a-8c6737b341f0" />

i feel like this path is wrong, but i can't come up with better one either?

does someone converting text to audio wants it to become txt2speech?

or perhaps map unicode to wav? since the max Unicode value is 1,114,111 (U+10FFFF), we can map it directly into the audio bit-depth.

p2r3:
There's currently only routes for rendering text visually, not as sound. The browser's TTS API is likely a good call for this kinda thing, assuming it can be used that way.

I'll reformat the issue a bit, because a question doesn't really constitute an issue.

Bee4Queer:
I can see about implementing this.

Bee4Queer:
Looks like the browser TTS API exclusively supports speaking text, not turning it into a file/data. I'm looking at [mespeak.js](https://www.masswerk.at/mespeak), though it seems old (and doesn't have a git repo, though it is GPL). Last update is 2020-04-23.

Bee4Queer:
There are definitely better options mespeak sucks for this usecase

Bee4Queer:
Oddly, unless I go into advanced mode and specifically choose the TTS provider for text/plain->audio/wav, it still tries to do text->PNG->wav, which should be impossible with BFS.

Bee4Queer:
Actually, I think I know why. It's getting a weird CORS error from trying to download a non-existent file on a CDN (don't know why it is trying to, but whatever) and it returning a 404 with an HTML document.

Bee4Queer:
Okay apologies but I cannot get this working, I'm happy to talk through what I've found overall if anyone is interested in getting this working

matan-h:
Compare between web TTS models I tried

- [kokoro](https://github.com/hexgrad/kokoro) - ~80MB model, good quality. make the tab unresponsive
- [piper](https://github.com/Mintplex-Labs/piper-tts-web) - ~25MB model, fairly good quality, however take too much time to convert most things. and take one full minute to convert 230 words.
- [Xenova/speecht5_tts](https://huggingface.co/Xenova/speecht5_tts) - ~80MB model - hard to config right so you can understand the output

ConnorTippets:
Have we considered the opposite, though? Converting text to audio and then converting back to text should produce the original text.

Bee4Queer:
[...]

PR #301, opened 2026-02-18
Add espeakng TTS

zipsegv:
Adds a TTS via [espeakng.js](https://github.com/steveseguin/espeakng.js/) (this PR uses a slight fork that makes it compatible with ES modules). It's not a very good TTS but it works, I guess.

This addresses issue #180.


WFlyToTheSky:
this probably isn't the best place to ask, but how you do the checks thing?

zipsegv:
> this probably isn't the best place to ask, but how you do the checks thing?

it does it automatically

WFlyToTheSky:
ok

PR #302, opened 2026-02-18
Updated mcmap to support to/from png as well as grid maps

toodu:
Updated the mcmapHandler to support PNG <-> Minecraft Map as well as PNG -> Grid Minecraft Maps (multiple maps for one picture).

### DIFF
--- a/src/handlers/mcmap.ts
+++ b/src/handlers/mcmap.ts
@@ -242,22 +242,16 @@ class mcMapHandler implements FormatHandler {
                         const width = NBT.isTag<NBT.IntTag>(mapdata["width"]) ? mapdata["width"].valueOf() : DEFAULT_WIDTH;
                         const height = NBT.isTag<NBT.IntTag>(mapdata["height"]) ? mapdata["height"].valueOf() : DEFAULT_HEIGHT;
                         const colors = NBT.isTag<NBT.ByteArrayTag>(mapdata["colors"]) ? new Uint8Array(mapdata["colors"]) : new Uint8Array([]);
-                        const bytes = new Uint8Array(map2rgba(colors, width, height));
+                        const rgba = map2rgba(colors, width, height);
 
-                        const blob = new Blob([file.bytes as BlobPart], { type: inputFormat.mime });
+                        this.#canvas.width = 128
+                        this.#canvas.height = 128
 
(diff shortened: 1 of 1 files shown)


WFlyToTheSky:
how do you chose the size of the gridmaps from the frontend? is it just hardcoded or is it based off of the image resolution?

toodu:
> how do you chose the size of the gridmaps from the frontend? is it just hardcoded or is it based off of the image resolution?

Right now I do not think it is possible to add settings so it is hard-coded to be the image resolution, rounded up to a multiple of 128

so for example an 128x128 image should be 1x1 map while a 230x128 image is should be stretched to be 256x128 (2x1 map grid)

The image used above is 480x480 so it is being stretched to 512x512 (4x4 map grid)

src/handlers/mcmap.ts
```diff
@@ -79,7 +82,10 @@ class mcMapHandler implements FormatHandler {
     public ready: boolean = false;
 
     async init() {
+
         this.supportedFormats = [
+            CommonFormats.PNG.supported("png", true, false, true),
+            CommonFormats.PNG.supported("png", false, true, false),
```
p2r3: Why is PNG listed twice? And with very conflicting settings?

It should just be `CommonFormats.PNG.supported("png", true, true, false)` if you wanted input/output support while indicating that it's a lossy conversion (which it is)

package.json
```diff
@@ -28,6 +28,7 @@
     "@types/three": "^0.182.0",
     "buffer": "^6.0.3",
     "imagetracer": "^0.2.2",
+    "jimp": "1.6.0",
```
p2r3: I really don't think we should be including yet another library for image processing.

For _most_ things, we already have the JavaScript canvas built into the browser. See `canvasToBlob.ts` for exmples on how to load a file onto it.

If you really need something more advanced (which I don't think this does), you may also use ImageMagick. But, again, since it seems you're just reading/writing pixel colors, the canvas is perfect for this. It can read/write PNGs just fine.

src/handlers/mcmap.ts
toodu: I incorrectly thought that since MAP->PNG is lossless that both should have been used, I have changed this now

PR #308, opened 2026-02-19
enabled input conversion for UTF-8 without BOM and UTF-8 with BOM formats

SPLATPLAYS:


millennIumAMbiguity:
Not necessary as the default reader handles both cases.
UTF8 with and without BOM is standardized in all text readers.

PR #316, opened 2026-02-19
Fix text→image conversion and resolve missing module

ashuwhy:
### 1. Text → PNG/JPEG "Canvas output failed"
- **Cause:** For empty or blank text, the canvas width/height could be 0, so `canvas.toBlob()` returned `null`.
- **Changes in `canvasToBlob.ts`:**
  - Set `ctx.font` before `measureText()` so layout and size are correct.
  - Enforce minimum canvas size (1×1) when rendering text so the canvas is never zero-size.
  - Pass explicit quality for JPEG output (`0.92`) when calling `toBlob(..., "image/jpeg", quality)`.

### DIFF
--- a/src/handlers/canvasToBlob.ts
+++ b/src/handlers/canvasToBlob.ts
@@ -42,20 +42,22 @@ class canvasToBlobHandler implements FormatHandler {
       if (inputFormat.mime === "text/plain") {
 
         const font = "48px sans-serif";
-        const fontSize = parseInt(font);
+        const fontSize = parseInt(font, 10);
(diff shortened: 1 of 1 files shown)


zipsegv:
> Cause: canvasToBlob imports imageToText and rgbaToGrayscale from that path, but the file did not exist.

this module exists, though? did you forget to run `git submodule update --init --recursive`?

ashuwhy:
Ya right i ran it, it's not in the commit i forgot to edit the comment out

PR #319, opened 2026-02-19
Add plain text to Go source code converter

SladkyCitron:
Added a plain text to Go source code converter (similar to text => batch).

It converts text like this:
```
Hello world!
the quick brown fox jumps over the lazy dog.
"quotes"
'single quotes'
```

...into Go code like this:
```go
package main

import "fmt"

func main() {
	fmt.Println(`Hello world!`)
	fmt.Println(`the quick brown fox jumps over the lazy dog.`)
	fmt.Println(`"quotes"`)
	fmt.Println(`'single quotes'`)
}
```

### DIFF
--- a/src/handlers/textToGo.ts
+++ b/src/handlers/textToGo.ts
@@ -13,7 +13,7 @@ class textToGoHandler implements FormatHandler {
       from: false,
       to: true,
       internal: "go",
-      category: "text",
+      category: "code",
       lossless: true,
     }
   ];


matan-h:
Sound nice, two questions:
1. Why not just write one `fmt.Println` for the entire file?
2. Currently the code does not escape `` ` ``, is it intentional  ?

SladkyCitron:
1. I did consider just writing a single `fmt.Println` but I decided for the multiple ones because I feel like it looks cleaner. I can switch to a single call if you want.
2. Escaping backticks isn’t currently handled, that isn’t intentional. I just wasn't able to find a good way to escape them.

SladkyCitron:
Update: switched to a single `fmt.Println` call. The example from earlier should now convert to this:
```go
package main

import "fmt"

func main() {
	fmt.Println(`Hello world!
the quick brown fox jumps over the lazy dog.
"quotes"
'single quotes'`)
}
```

p2r3:
> Escaping backticks isn’t currently handled, that isn’t intentional. I just wasn't able to find a good way to escape them.

Please do. When handling generated code, these things are important.

SladkyCitron:
Added proper backtick handling. I still couldn't find a good way to escape them so I'm using concatenation instead.

Here's how the generated code looks like:
```go
package main

import "fmt"

func main() {
	fmt.Println(`backtick` + "`" + `test`)
}
```

p2r3:
The usual way to sanitize certain characters is to first sanitize the escape character (e.g. `\` becomes `\\`), and then append the escape character to the target character (e.g. `` ` `` becomes `` \` ``).

I assume that's possible in Go? Would be cleaner than concatenation.

SladkyCitron:
> I assume that's possible in Go? Would be cleaner than concatenation.

In Go, raw string literals (backticks) don't support escape sequences at all, so there’s actually no way to escape a backtick inside them. The only way to include one is via concatenation and switching to a quoted string just for the backtick, which is what I’m doing here.

I'm intentionally using raw string literals instead of the regular quoted ones because they let me preserve newlines and other characters without having to escape everything (e.g. `\n`, `\"`, `\\`) which keeps the generated code much cleaner and readable. Backticks are the only character that require special handling here.

src/handlers/textToGo.ts
```diff
@@ -0,0 +1,59 @@
...
+    CommonFormats.TEXT.supported("txt", true, false, true),
+    {
+      name: "Go Source File",
+      format: "go",
+      extension: "go",
+      mime: "text/x-go",
+      from: false,
+      to: true,
+      internal: "go",
+      category: "text",
```
p2r3: This should probably use the "code" category.

SladkyCitron: Good catch, updated.

PR #329, opened 2026-02-19
Implement sgxmiscHandler for SGX format conversion

BLACKBIRT007:
sgx to json and .txt and back for the stl to sgx and sgx to stl converter


p2r3:
This does not count as converting a file. You're just dumping the data straight into the new container without parsing it.

PR #335, opened 2026-02-19
add valve material (.vmt) support

iliyalb:
essentially a wrapper using vtf


p2r3:
No, no, no! Handlers should not be calling frontend code!! It's possible we'll have some kind of clean interface for frontend-to-backend communication some day, but this still would _not_ be the way to implement this format.

A VMT file is, effectively, a metadata file. It only makes sense to treat it as such, i.e., converting it to JSON or similar. If the user wants to convert a VTF, they'll input a VTF.

Issue #333, opened 2026-02-19
[File format] MIDI files


iAmInActions:
**Describe the file format**
Musical Instrument Digital Interface ([/ˈmɪdi/](https://en.wikipedia.org/wiki/Help:IPA/English); MIDI) is an American-Japanese [technical standard](https://en.wikipedia.org/wiki/Technical_standard) that describes a [communication protocol](https://en.wikipedia.org/wiki/Communication_protocol), [digital interface](https://en.wikipedia.org/wiki/Digital_electronics), and [electrical connectors](https://en.wikipedia.org/wiki/Electrical_connector) that connect a wide variety of [electronic musical instruments](https://en.wikipedia.org/wiki/Electronic_musical_instrument), [computers](https://en.wikipedia.org/wiki/Computer), and related audio devices for playing, editing, and recording music

*source: [Wikipedia - MIDI](https://en.wikipedia.org/wiki/MIDI)*

**Describe the expected behavior**
The conversion should support both going from a MIDI file to audio using a small soundfont, for example [Tim6MBGM](https://musical-artifacts.com/artifacts/7293) which is public domain and around 6 MB, as well as going from audio to MIDI using transformation and density mapping. As a side effect of my intended way to implement this, it would also support going from image to midi as the bright spots would map onto notes and thus create a piano roll representing the images content. (so "convert to midi" input would most likely be PNG format)

**Provide reference implementations**
- [MIDI Spec](https://midimusic.github.io/tech/midispec.html)
- [Soundfont](https://mrtenz.github.io/soundfont2/getting-started/soundfont2-structure.html)
- [MidiWriterJS](https://github.com/grimmdude/MidiWriterJS)

**PLEASE NOTE:** Simply listing a file format with no further justification beyond "I want this" will likely get your issue closed. There are a lot of file formats out there, and it can take hours to add support for just one. Just naming them isn't productive, so please try to be forthcoming with resources of your own.

**Do you plan on implementing this yourself?**
Yes, I already have some local tests for the general concept, currently haven't had the time to properly implement it into convert.to.it yet though. Making this issue for progress tracking once I start working on it.

**Additional context**
- [FluidSynth Emscripten](https://github.com/jet2jet/fluidsynth-emscripten) for MIDI -> Audio
- Already implemented Audio -> PNG route for baseline of Audio -> MIDI

p2r3:
[...]

PR #350, opened 2026-02-20
Add support for MIDI -> Audio | see #333 

iAmInActions:
See #333 for details

Groundwork:
- [x] fluidsynth
- [x] Custom MIDI file format handler
  - [x] Read file
  - [x] Write file
  - [x] Parse into list
  - [x] Convert into strings
  - [x] Convert strings into list
  - [x] Manually handle notes

Implemented handlers so far:
- [x] MIDI to Audio (fluidsynth)
- [x] better MIDI to image (own implementation)
- [x] MIDI to text (own implementation)
  - [x] txtMIDI into MIDI
  - [x] grub init into MIDI
  - [x] nokia into MIDI
  - [ ] ~~generic text (parse ASCII values as notes/duration)~~\*\*
- [x] Text to MIDI\* (own implementation)
- [x] Image to MIDI (own implementation)
  - [x] Audio to MIDI (existing wav -> png path)

Additional details / things I learned along the way:
- The import path being used can lead to re-declaration attempts, need to be careful to work around that.
- Most JS MIDI parsing libraries are either for webserial to use keyboards or just suck, will write my own in the scheme of this handler.

Edit: please do not merge this yet, I want to finish this first. Tracking here.

\*Text to MIDI currently only support text in the format of midi to text, still need to add support for detecting and properly handling generic text, grub init tunes and nokia tunes

\*\*Scrapped to keep existing espeak text to audio as its more useful than interpreting text into notes.

### DIFF
--- a/src/handlers/midi.ts
+++ b/src/handlers/midi.ts
@@ -87,7 +87,7 @@ export class midiCodecHandler implements FormatHandler {
       { name: "RTTTL",         format: "rtttl",  extension: "rtttl",  mime: "audio/rtttl",  from: true,  to: true,  internal: "rtttl", category: "text",  lossless: false },
       { name: "NokRing",       format: "rtttl",  extension: "nokring",mime: "audio/rtttl",  from: true,  to: false, internal: "rtttl", category: "text",  lossless: false },
       { name: "GRUB Init Tune",format: "grub",   extension: "grub",   mime: "text/plain",   from: true,  to: true,  internal: "grub",  category: "text",  lossless: false },
-      { name: "Plain Text",    format: "txt",    extension: "txt",    mime: "text/plain",   from: true,  to: true,  internal: "txt",   category: "text",  lossless: true },
+      { name: "Plain Text",    format: "text",    extension: "txt",    mime: "text/plain",   from: true,  to: true,  internal: "txt",   category: "text",  lossless: true },
       // PNG spectrogram -> MIDI (matches meyda's internal="image" so routing picks
       // up the audio->png->mid path automatically)
       { name: "PNG",           format: "png",    extension: "png",    mime: "image/png",    from: true,  to: true,  internal: "image", category: "image", lossless: false },


p2r3:
> Edit: please do not merge this yet, I want to finish this first. Tracking here.

I'll convert this to a draft until it's finished then.

p2r3:
Hey, your `txt->wav` "edge case" made me realize we have a shortcoming in our routing algorithm, where files with different meanings but similar MIME types couldn't be reasonably distinguished.

I've adjusted the graph traversal to take into account the `format` string on the first and last node, so if you wish, you may revert to having that `txt->wav` edge, as long as the format string is something other than `text` (e.g. `text-notes`).

p2r3:
To clarify about my last change, this just means that when a user is selecting their input/output formats, they'd have the option to select between `TEXT` (which would be the default), and `TEXT-NOTES`, for example. If the format is in the middle of a routing path, then either can be picked depending on their weight in that specific scenario.

iAmInActions:
> To clarify about my last change, this just means that when a user is selecting their input/output formats, they'd have the option to select between `TEXT` (which would be the default), and `TEXT-NOTES`, for example. If the format is in the middle of a routing path, then either can be picked depending on their weight in that specific scenario.

sounds reasonable although after further thought I still dont really see a use in having the "turn ascii character pairs into note+duration" parsing when we already have actually human read- and writeable formats like rtttl and grub init supported by this. speaking the text from a normal txt just seems like the better option even if both are available. no one will choose the second. I think its better if i focus on getting things like image to audio and audio to image done now so this can be merged.

p2r3:
Sure 👍

iAmInActions:
You fine with if i cap the meyda output at 24 kHz? audio formats are oversampled to compensate for rounding noise, human hearing only goes up to around 20 kHz so going any higher in the image output wastes resolution that could otherwise more accurately reconstruct the audio from compressed images or turn it into midi.

edit: since you mentioned not being too familiar with this in your video, if you transform the audio into its elements, you get rid of these rounding issues so you only need the "real" spectrum and not the soundcards sampling spectrum. its the same idea as an oscilloscope with 40 megasamples only being able to display up to 20 MHz properly

iAmInActions:
Correction, hearing range is up to 20k although commonly ~17k is the limit so ill say for the best result in terms of quality, 17 is a good choice

p2r3:
Sure, from what I can tell, it seems like you know what you're talking about in regards to audio. Tested, seems to work, checking out the code now.

PR #355, opened 2026-02-20
font support (ttf, otf, woff, woff2 to svg, otf, woff2)

alfuwu:
adds (basic) support for converting popular font files

unfortunately, doesn't really work converting something else (like a png) *to* a font file (the medium the files were connected to the main conversion formats was SVG by way of SVG fonts, which as far as i can see, nothing else converts to an SVG font, and thus nothing outside of other font formats can be converted between font formats). maybe you could do some fancy method of rendering (and then detecting) the glyphs directly in the SVG as a form of image atlas? but even then, vital metadata (such as what glyph an SVG element is representing) would still need to be included, which, if you converted a PNG to an SVG, wouldn't exist.

additionally, due to limitations of the primary library used here (opentype.js), outputting ttf and woff files isn't supported

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "ts-flp": "^1.0.3",
     "vite-plugin-static-copy": "^3.1.6",
     "wavefile": "^11.0.0",
-    "woff2-encoder": "^2.0.0"
+    "woff2-encoder": "^2.0.0",
     "js-synthesizer": "^1.11.0"
   }
 }


src/handlers/font.ts
```diff
@@ -0,0 +1,260 @@
+import { Font, Glyph, Path, parse } from "opentype.js";
+import { SVGPathData } from "svg-pathdata";
+import { compress, decompress} from 'woff2-encoder';
```
alfuwu, line 3: should be `import { compress, decompress } from "woff2-encoder";`

src/handlers/font.ts
```diff
@@ -0,0 +1,260 @@
...
+  return { bytes, name };
+}
+
+async function sfntToWoff2(inputFile: FileData): Promise<FileData> {
+  const font = parse(inputFile.bytes.buffer);
+  const sfnt = new Uint8Array(font.toArrayBuffer());
+
+  const bytes = await compress(sfnt);
+
+  const name =inputFile.name.split(".")[0] + ".woff2";
```
alfuwu, line 204: should  be `const name = inputFile.name.split(".")[0] + ".woff2";`

src/normalizeMimeType.ts
```diff
@@ -17,6 +17,13 @@ function normalizeMimeType (mime: string) {
     case "application/x-lha": return "application/x-lzh-compressed";
     case "application/x-lzh": return "application/x-lzh-compressed";
     case "audio/x-flac": return "audio/flac";
+    case "application/font-sfnt": return "font/ttf";
+    case "application/x-font-ttf": return "font/ttf"; // both TTF & OTF
```
alfuwu, line 21: the `// both TTF & OTF comment` was put on the wrong line somehow, it sohuld be on `application/font-sfnt`, not `application/x-font-ttf`

PR #356, opened 2026-02-20
added txtToPy converter

wenxigoh:


wenxigoh:
it just prints out what the txt file has

src/handlers/index.ts
```diff
@@ -36,6 +36,7 @@ import espeakngHandler from "./espeakng.js"
 import textToShellHandler from "./texttoshell.ts";
 import batchHandler from "./batch.ts";
 import bsorHandler from "./bsor.ts";
+import TxtToPyHandler from './textToPy.ts';
```
p2r3: Handler imports/exports should be in camelCase.

src/handlers/index.ts
```diff
@@ -76,5 +77,6 @@ try { handlers.push(new espeakngHandler()) } catch (_) { };
 try { handlers.push(new textToShellHandler()) } catch (_) { };
 try { handlers.push(new batchHandler()) } catch (_) { };
 try { handlers.push(new bsorHandler()) } catch (_) { };
+try { handlers.push(new TxtToPyHandler()) } catch (_) { };
```
p2r3: I think the placement of this will cause the format entry to read "Python turtle program" due to the other Python handler higher up.

I suggest creating a new CommonFormats entry for Python files (if there isn't one already) and editing both this handler and `pyTurtleHandler` to use it. This would fix the naming issue and would also standardize the Python output format, letting the routing algorithm pick whichever one makes the most sense in any given scenario.

wenxigoh: That's true, without turning on advanced mode the website only shows the "Python turtle program" option.

wenxigoh:
I don't think I used CommonFormats correctly

src/CommonFormats.ts
```diff
@@ -110,6 +110,13 @@ const CommonFormats = {
         "text/windows-batch",
         ["text"]
     ),
+    PYTHON: new FormatDefinition(
+        "Python Script",
+        "py",
+        "py",
+        "text/x-python",
+        [Category.TEXT, Category.IMAGE]
```
p2r3: Seems correct overall, but Python code is definitely not in the "text" or "image" category 😅

You can create a new enum for `Category.CODE`, or just use "code" as a literal string. The former is obviously preferred.

wenxigoh:
That should be it

ffogells (approved):

PR #365, opened 2026-02-20
Add Electron Support

### LINKED ISSUE #360: [Feature] Electron App support
**Is your feature request related to a problem? Please describe.**
The website, while convenient, can only be used online. While there are ways to turn it into a web archive, it is overkill and requires significant overhead, depending on the format. <sub>(I'm looking at you, WARC and ReplayWeb.page)</sub>

**Describe the solution you'd like**
Since this is based on Node.js and Vite, it's easy to create an application that runs offline. Essentially:
- This would make the program better and easier to use
- It further separates the distinction between this and those online, server-based converters
- It just sounds cool to have a super-converter on hand that does everything, regardless of internet connection

**Describe alternatives you've considered**
It would be better to have a more native, CLI program. However, given the trade-offs, I think the Electron option will still be great.

**Additional context**
I am working on making this possible. Still in the draft stages, but you can look at the `electron-shenanigans` branch in my fork.

exurd:
This PR adds Electron support. Closes #360 if merged.

Things to do:
- [x] Rename app name to `p2r3-convert`
- [x] Add author and description to program
- [x] Open GitHub page in default browser instead of denying any access
- [x] Fix Windows NSIS installer
  - hmmm... fixed. the github action artifact installed normally. (maybe it was from the app rename? or just pure windows sandbox insanity? idk anymore)
- [x] Setup GitHub actions to create artefacts

### DIFF
--- a/src/electron.cjs
+++ b/src/electron.cjs
@@ -1,6 +1,8 @@
 const { app, BrowserWindow, protocol, net, shell } = require('electron');
-const { URL } = require('node:url');
+const { URL, pathToFileURL } = require('node:url');
 const path = require('path');
+
+const DIST_PATH = path.join(__dirname, '..', 'dist');
 
 protocol.registerSchemesAsPrivileged([
   {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,20 +1,51 @@
 {
-  "name": "convert",
+  "name": "p2r3-convert",
+  "productName": "Convert to it!",
+  "author": "PortalRunner",
+  "description": "Truly universal online file converter",
```
exurd: Should the "online" part be in the description? Since this PR allows the converter to be offline as well, it should probably reflect that. Your call, though.

src/electron.cjs
```diff
@@ -0,0 +1,80 @@
...
+    let urlPath = request.url.replace('app://-/', '');
+    if (urlPath.includes('convert/')) {
+      urlPath = urlPath.replace('convert/', '');
+    }
+
+    // Remove query parameters or hashes just in case Vite adds them (e.g. ?worker_file)
+    urlPath = urlPath.split('?')[0].split('#')[0];
+
+    // Fetch the local file in dist
+    const filePath = path.join(__dirname, '..', 'dist', urlPath);
```
p2r3: `urlPath` should _probably_ be checked for path traversal here. I'm not sure how much sandboxing Electron does, but better safe than sorry.

package.json
p2r3: You can do "browser-based" instead of "online" here.

src/electron.cjs
```diff
@@ -0,0 +1,80 @@
+const { app, BrowserWindow, protocol, net, shell } = require('electron');
```
p2r3, line 1: Mega uber tiny nitpick, but I'd prefer if this file was just `electron.cjs` instead if not too much of a hassle. There's only one file for Electron stuff, no need to overdeclare its purpose.

exurd: Resolved in 40bd773.

package.json
exurd: Resolved in e439be6.

src/electron.cjs
exurd: The URL checks should be much better in e5a255a

p2r3:
brother you better pray this works

p2r3:
didn't pray hard enough

zardoy:
Some feedback (if you dont mind:)

- why `p2r3-convert` instead of `convert-to-it` ? It would sound more familiar to the domain name which imo makes sense
- it would be nice to add meta information publishing ([example](https://github.com/zardoy/aqua-player/tree/release-info)) so info like version can be fetched from github without cors issues (if you're gonna plan to set up a version link on some website in the future)
- There are still yet many options missing in electron builder config like icons, mac entitiles or external ffmpeg bundle (yes would be better to ship with full native binaries since Chromium bindings can only play some codecs), so to make it cleaner IMO its better to split into its own file `electron-builder-config.js`

PR #369, opened 2026-02-20
Added ZSTD support

chippytech:
I added ZSTD support via Webassembly.


p2r3:
Hm, outputting some nondescript binary file doesn't really make much sense here. If you had a ZSTD compressed DOCX and wanted to turn that into a PDF, this wouldn't really let you do that.

I'd almost like to treat this as a "meta" handler, in the sense that it doesn't really fit in the regular conversion graph at all. There's underlying data of its own type that we can _probably_ handle with something else, but not before it's passed through this.

My thoughts on how this should be implemented are very divided, but it certainly seems like it needs more work and consideration beyond simply decompressing to a binary blob.

i-hate-trustedinstaller:
are you still working on this?

Issue #368, opened 2026-02-20
[File format] JSON to BSON conversion


piotrstenke:
**Describe the file format**
BSON is a binary data exchange format based on JSON. It is used for transfering data in e.g. NoSql databases (MongoDB).

**Describe the expected behavior**
Simple, lossless conversion from JSON to object in Java Script to BSON.

**Provide reference implementations**
[npm](https://www.npmjs.com/package/bson)

**Do you plan on implementing this yourself?**
Yes

SPLATPLAYS:
Should this issue be closed?

PR #371, opened 2026-02-20
Add conversion from JSON to BSON and vice versa

piotrstenke:
#368 

Can be verified e.g. here: https://jsontotable.org/bson-to-json


p2r3:
I'd prefer if this was merged into one handler instead of two separate ones. You can simply check the direction of the conversion with a couple of if statements. That would reduce application and cache size and routing complexity.

PR #377, opened 2026-02-21
CUR <-> ICO support and ANI <-> CUR support

omega-fallon:
Adds support for converting to and fro .cur and .ico files, as well as extracting the first frame of .ani files and outputting them as .cur's!

### DIFF
--- a/src/handlers/aseprite.ts
+++ b/src/handlers/aseprite.ts
@@ -1,0 +1,477 @@
+import pako from "pako";
+import CommonFormats from "src/CommonFormats.ts";
+import type { FileData, FileFormat, FormatHandler } from "../FormatHandler.ts";
+
+const ASEPRITE_HEADER_MAGIC = 0xA5E0;
+const ASEPRITE_FRAME_MAGIC = 0xF1FA;
+const CHUNK_TYPE_LAYER = 0x2004;
+const CHUNK_TYPE_CEL = 0x2005;
+const CHUNK_TYPE_PALETTE = 0x2019;
+const CHUNK_TYPE_PALETTE_LEGACY_8BIT = 0x0004;
+const CHUNK_TYPE_PALETTE_LEGACY_6BIT = 0x0011;
+
+interface LayerInfo {
+  visible: boolean;
+  opacity: number;
+  blendMode: number;
+}
+
+interface CelImage {
+  width: number;
+  height: number;
+  x: number;
+  y: number;
+  layerIndex: number;
+  opacity: number;
+  pixels: Uint8Array;
+}
+
+interface ParsedAseprite {
+  width: number;
+  height: number;
+  pixels: Uint8ClampedArray;
+}
+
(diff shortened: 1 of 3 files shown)


src/handlers/curani.ts
```diff
@@ -0,0 +1,167 @@
...
+                        let i = 0;
+
+                        // Finds where the first ICO header is
+                        while (header_hook === 0) {
+                            if (new_file_bytes[i] == 0x28 && new_file_bytes[i+1] == 0x0 && new_file_bytes[i+2] == 0x0 && new_file_bytes[i+3] == 0x0 && new_file_bytes[i+4] == 0x20 && new_file_bytes[i+5] == 0x0 && new_file_bytes[i+6] == 0x0 && new_file_bytes[i+7] == 0x0 && new_file_bytes[i+8] == 0x40) {
+                                header_hook = i;
+                            }
+                            
+                            if (i > new_file_bytes.length) {
+                                throw new Error("FUCK FUCK FUCK.");
```
p2r3: Not a particularly descriptive error.

src/handlers/curani.ts
```diff
@@ -0,0 +1,167 @@
...
+                        while (counter < images_present) {
+                            // color planes
+                            new_file_bytes[10+(counter*16)] = 0;
+                            new_file_bytes[11+(counter*16)] = 0;
+                            // bits per pixel
+                            new_file_bytes[12+(counter*16)] = 0;
+                            new_file_bytes[13+(counter*16)] = 0;
+                            counter += 1;
+                        }
+                        //throw new Error("NEEDS TO BE IMPLEMENTED.");
```
p2r3: Leftover comment?

src/handlers/curani.ts
```diff
@@ -0,0 +1,167 @@
...
+                }
+
+                outputFiles.push({
+                    name: file.name.split(".").slice(0, -1).join(".") + "." + outputFormat.extension,
+                    bytes: new_file_bytes
+                })
+                return outputFiles;
+            }
+            catch (e) {
+                console.error(e);
```
p2r3: You should probably actually throw the errors (or just don't catch them) instead of returning an empty output. The upstream code has its own error handling, and an error is the cleanest way to signal that something isn't supported.

src/handlers/curani.ts
```diff
@@ -0,0 +1,188 @@
...
+                to: true,
+                internal: "ani",
+                category: "image",
+                lossless: false,
+            },
+            {
+                name: "Microsoft Windows CUR",
+                format: "cur",
+                extension: "cur",
+                mime: "image/vnd.microsoft.cursor",
```
p2r3, line 32: If you merge with upstream, you should be able to now change this back to the correct MIME type (image/vnd.microsoft.icon) while still maintaining separation. (Introduced in #383)

pompom454:
Looking good!

PR #378, opened 2026-02-21
fix: correctly compare input and output formats during file download

ConnorTippets:


ConnorTippets:
oh god fucking dammit vscode
my formatter changed like everything without my knowledge or consent

Issue #322, opened 2026-02-19
[File format] .schematic or .schem (Minecraft)


OwlWorksInnovations:
**Describe the file format**
The .litematic file format is used by the Litematica mod for Minecraft (Java Edition). It stores 3D block structures, including block states, tile entity data, and metadata such as region name and dimensions. Files are NBT (Named Binary Tag) encoded and compressed with GZIP. It is used primarily by Minecraft players to save and share building blueprints.

**Describe the expected behavior**
Convert .litematic files to/from .schem (WorldEdit schematic format) or .nbt (Minecraft structure format). The conversion should preserve block types, positions, and block states as accurately as possible.

**Provide reference implementations**
[Litematica source code (LGPL-3.0)](https://github.com/maruohon/litematica) - reference for parsing the format
[NBT.js (MIT)](https://github.com/sjmulder/nbt-js)

**PLEASE NOTE:** Simply listing a file format with no further justification beyond "I want this" will likely get your issue closed. There are a lot of file formats out there, and it can take hours to add support for just one. Just naming them isn't productive, so please try to be forthcoming with resources of your own.

**Do you plan on implementing this yourself?**
No

**Additional context**
The format is widely used in the Minecraft community, particularly among survival players. Many builds are shared exclusively in .litematic format. And a popular mod named "Create" uses .nbt for a item named "Schematic Cannon" to build schematics automatically.

Larson-Logan:
I do have this tool created already, however, it was created in java and I dont know how easy it would be to implement this here. If you need it heres a link to the repo: https://github.com/Larson-Logan/SchemConvert/releases/tag/v1.3.1

PR #379, opened 2026-02-21
Added support for .litematic, .schem, .schematic formats used it minecraft.

Larson-Logan:
Resolves https://github.com/p2r3/convert/issues/322

### DIFF
--- a/src/handlers/mcSchematicHandler.ts
+++ b/src/handlers/mcSchematicHandler.ts
@@ -2,8 +2,8 @@ import type { FileData, FileFormat, FormatHandler } from "src/FormatHandler";
 import * as NBT from "nbtify";
 import { gunzipSync, gzipSync } from "fflate";
 
-class schematicConverter implements FormatHandler {
-    public name: string = "schematicConverter";
+class mcSchematicHandler implements FormatHandler {
+    public name: string = "mcSchematic";
     public supportedFormats?: FileFormat[];
     public ready: boolean = false;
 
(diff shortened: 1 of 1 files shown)


src/handlers/nbt.ts
```diff
@@ -22,6 +23,39 @@ class nbtHandler implements FormatHandler {
...
+                name: "Litematica Schematic",
+                format: "LITEMATIC",
+                extension: "litematic",
+                mime: "application/x-minecraft-litematic",
+                from: true,
+                to: false,
+                internal: "litematic",
+                category: "data",
+                lossless: true
+            },
```
p2r3: - The format string should be in lowercase. This includes NBT as well, it seems I hadn't caught that before.
- Correct me if I'm wrong, but it seems like this handler does not support schematic input. If so, shouldn't the `from` fields be false? And therefore `litematic` removed due to redundancy?

src/handlers/mcSchematicHandler.ts
```diff
@@ -0,0 +1,387 @@
...
+                name: "Named Binary Tag",
+                format: "NBT",
+                extension: "nbt",
+                mime: "application/x-minecraft-nbt",
+                from: false,
+                to: true,
+                internal: "nbt",
+                category: "data",
+                lossless: true
+            }
```
p2r3, line 56: Same as before, `format` should be lowercase for all.

src/handlers/nbt.ts
Larson-Logan: I was able to just remove all three of those, as the schematicConverter.ts handles it already.

PR #380, opened 2026-02-21
Correctly compare input and output formats

ConnorTippets:
PR title is garbage ik


exurd:
Do you know any issues that this PR would close/fix? I'm trying to figure out how adding that check will be helpful.

p2r3:
Superseded by #383

p2r3:
@exurd this was motivated by a conversation in Discord, until I realized that this is inefficient to solve the particular problem we encountered.

The issue is that some files share one and the same MIME type, while being quite different in practice. Our current routing algorithm fails to take this into account.

_This_ PR in particular was to fix an edge case (didn't end up working).

PR #381, opened 2026-02-21
feat: add ICNS support for macOS icon conversion

Aspenini:
- add .icns as a supported input/output format
- add ICNS encoding/decoding to the conversion pipeline


src/handlers/icns.ts
```diff
@@ -0,0 +1,235 @@
...
+class icnsHandler implements FormatHandler {
+
+  public name: string = "icns";
+  public ready: boolean = false;
+  public supportedFormats: FileFormat[] = [
+    CommonFormats.PNG.supported("png", true, true, true),
+    ICNS_FORMAT.builder("icns")
+      .allowFrom(true)
+      .allowTo(true)
+      .markLossless(true)
```
p2r3: Are we sure these are lossless? I'm not familiar with `icns`, but it seems like it's limited to a finite set of resolutions, which would imply lossy scaling,

PR #386, opened 2026-02-21
Add .har (HTTP Archive) file support

Xeltalliv:
This pull request adds a support for [HAR format](https://en.wikipedia.org/wiki/HAR_(file_format)) and 2 handlers for it:
- har to zip conversion, which is practically useful for downloading every file of a webpage in it's original form
- har to json renaming


p2r3:
It seems like this handler would benefit greatly from outputting a single HTML file with all of the media loaded by the browser embedded via base64. I'm not entirely familiar with this format, so I have to ask:
- Is this possible or am I misunderstanding how the file works?
- Would you be interested in doing that for this PR, or would you rather I merge as-is?

Xeltalliv:
> Is this possible or am I misunderstanding how the file works?

For simple static webpages it should definitely be possible (to parse html and inline all the referenced urls as data uris, as long as their content was also captured).

> Would you be interested in doing that for this PR, or would you rather I merge as-is?

So far I don't feel like working on it, so I'd rather this get merged as is.

p2r3:
This doesn't build:
> src/handlers/har.ts(54,70): error TS18048: 'fileName' is possibly 'undefined'.

Please test your pull requests with `bun run build`

PR #389, opened 2026-02-21
feat: add system/light/dark theme toggle with persisted preference

Aspenini:
- add `System / Light / Dark` theme selector
- save user choice in `localStorage`


p2r3:
Sorry, we won't be making any frontend changes right now in order to avoid causing conflicts with #73.

Issue #382, opened 2026-02-21
[File format] MathML


bas1licate:
**Describe the file format**
MathML is an XML-based **Math**ematical **M**arkup **L**anguage used for displaying mathematical equations in XML-based file formats. it's most common use is embedded in HTML documents via the <math> tag.

**Describe the expected behavior**
MathML (.mml) (application/mathml+xml) can easily be converted to HTML by wrapping it in a <math> tag if everything isn't already, which can be done with parsing on the more simple side, but converting to it from HTML would be a more difficult problem to solve. it could also potentially be converted to/from LaTeX, although you would have to write a parser for that.

**Provide reference implementations**
specification is available at https://w3.org/TR/MathML. if/when a new version that is significant enough releases, i may open an issue to update any attached parsers if needed, although the MathML to HTML "parser" that checks if everything is inside a <math> tag and wraps whatever isn't in one likely wouldn't need updating after it's finished.

**Do you plan on implementing this yourself?**
I would try to implement basic capability myself (no-parsing method, which would then be replaced by the somewhat simple parsing if i could figure it out) if i knew how to implement a format. Once I figure out how, I will very likely try to implement it myself. (update: i tried. my implementation sucked so it was scrapped. there's always next time.)

**Additional context**
the idea for conversion to/from LaTeX came from it being a way to get similar mathematical equation capabilities to that in HTML. it shouldn't be too hard to implement converting from MathML to other formats. converting to MathML from other formats will likely mean a loss of formatting, unless a way is found to somehow preserve formatting. while typically inline, it also supports blocks via the display attribute.

p2r3:
> if i knew how to implement a format

Check out the readme, it has an example and a few key pointers to get you going.

bas1licate:
since i added in the "Once i figure out how" part, i've been working on something. i'll say when it's done

bas1licate:
mmlwrap is done. requesting pull if that's ok with you (although technically i won't request it until i'm sure it's ok with you or if enough time passes without a response)

bas1licate:
it's been enough time by now.

bas1licate:
#390 is up

bas1licate:
small update, since @matan-h told me about an implementation that should be very easy over in #390 , i think we can close this issue either now or when that pr is done.

waiting on feedback and/or green-light/go-ahead from you @p2r3

bas1licate:
[...]

PR #390, opened 2026-02-21
add MathML capability (using pandoc, and with a basic MathML to HTML converter as backup if that fails)

bas1licate:
basic MathML capability through embedding the entire thing inside an html <math> tag.

### DIFF
--- a/src/handlers/mmlwrap.ts
+++ b/src/handlers/mmlwrap.ts
@@ -22,7 +22,7 @@ class mmlwrapHandler implements FormatHandler {
       // Alternatively, if you need a custom format, define it like so:
       {
         name: "Mathematical Markup Language (MathML)",
-        format: "MathML",
+        format: "MathML (wrap)",
         extension: "mml",
         mime: "application/mathml+xml",
         from: true,


bas1licate:
(note: the secomd commit is a byproduct of syncing the fork)

matan-h:
isn't it better to use pandoc, since we already include that and it have good quality support for MathML?

urls:  
* https://leancrew.com/all-this/2025/05/mathml-with-pandoc/
* https://pandoc.org/demo/example33/3.6-math-rendering-in-html.html

bas1licate:
that actually sounds like a great idea for converting to MathML, which i initially expected to be harder to implement than converting from MathML due to not knowing of any existing solutions.
mathwrap.ts is a temporary file used to add basic MathML to HTML conversion, and i should mention that it is temporary until i can make a more proper one, which can be under the same or a different name.

bas1licate:
@matan-h if it concerns you i added adding pandoc to this.
(also the 5th commit, and any commits that say the same thing 2nd and 5th say, is yet again brought on by syncing fork)

matan-h:
sorry, I meant this project already have pandoc hander [here](https://github.com/p2r3/convert/blob/34b69a18f6c7603ea975a6684de49af047796fbf/src/handlers/pandoc.ts#L6) and the [pandoc.wasm](https://github.com/p2r3/convert/blob/34b69a18f6c7603ea975a6684de49af047796fbf/vite.config.js#L21) file. there is no need to re-add it as a submodule. I suggested you can use that, to better convert MathML to html and back with full spec support. I don't really understand MathML format, so feel free to ignore me :)

bas1licate:
@matan-h it should work, probably.
and if it does work, we can close #382 by the time this pr is done

bas1licate:
i was going to delete mmlwrap.ts, but instead i'll just keep it as a backup incase pandoc fails to convert.

p2r3:
???
This was a spectacular breakdown in communication.

According to the links posted by @matan-h, Pandoc does indeed seem to support MathML. And, according to those same links, it's an _option_ for the input, not a separate file format. _If it were_ a separate file format, it would already show up on the list. WASM Pandoc exposes all of the formats it supports automatically:
https://github.com/p2r3/convert/blob/2eaaa295a6ec86222e43ac031e9a7ec6362d6252/src/handlers/pandoc.ts#L166-L167 The `formatNames` and `formatExtensions` arrays are only there to assign these formats neat labels and appropriate file extensions.

That's why I'm incredibly confused by this commit - fa0d203faf793ab2aa8888c4214c6267c6f3dbec. Why did you think that giving it a name and extension would straight up implement it? Rather, how did you not notice that it _didn't_ implement it? Surely you would've tested your change??

Anyway, yes, I do believe that the correct solution is to implement MathML support through Pandoc. I don't really see much of a reason for this standalone handler to exist if the "elegant" variant is already built into the program, gated only by a flag and a bit of scaffolding.

I'm closing this in favor of an implementation that just uses Pandoc correctly.

bas1licate:
ok

PR #392, opened 2026-02-21
added german lang support and fixed some bugs

Ben-bit-code208:


eilon40:
I think it might be better to handle the languages using JavaScript instead of creating separate new HTML files, because any change in the HTML would require updating additional files as well.

p2r3:
We're currently not accepting any UI changes until #73 is finished.

Also, as @eilon40 said, we should first develop a proper localization system and _then_ localize. (I'm also not sure what "bugs" were fixed? Seems like you just made a few inconsequential config changes.)

PR #393, opened 2026-02-21
N64 ROM conversion (.n64/.z64/.v64, plus lossless .z64 to .png) and .bik Bink video input support

Aspenini:
- Added N64 ROM format support for `.z64`, `.n64`, and `.v64` with lossless conversion between each byte order.
- Added lossless `.z64 ↔ .png` conversion using literal pixel-channel data (no hidden metadata), with an opaque RGBA mapping to preserve exact round-trips.
- Added `.bik` Bink video input support by normalizing FFmpeg format metadata for reliable extension detection.


src/handlers/n64rom.ts
```diff
@@ -0,0 +1,282 @@
...
+      name: "N64 ROM Packed Image (PNG)",
+      format: "png",
+      extension: "png",
+      mime: PNG_MIME,
+      from: true,
+      to: true,
+      internal: "n64png",
+      category: ["image", "data"],
+      lossless: true
+    }
```
p2r3: This should just use the CommonFormats definition of a PNG. The custom name isn't going to be used anyway, other handlers with higher priority replace it for this MIME type.

p2r3:
In the future, please make one pull request per feature.

PR #395, opened 2026-02-21
feat: Support for Pâquerette: Down the Bunburrows levels

omega-fallon:
Added a handler which renders the custom level format of PDtB as PNGs.


p2r3:
This fails to build due to an undeclared variable. You can use `bun run build` to verify that your fork builds before PRing it

p2r3:
In the future, please do not touch code unrelated to your pull request. This has changes in the `ani` and `mcmap` handler which, while seemingly helpful, do not belong to this context.

PR #399, opened 2026-02-22
add mathml conversion using pandoc

ConnorTippets:
i believe i implemented this correctly? i mean, it functions.
i put that part in [normalizeMimeType.ts](https://github.com/ConnorTippets/convert/blob/ccd740ba1be52aa3ec387dbed3de038a714ee437/src/normalizeMimeType.ts#L30) because the existing code gets mimetype by file extension, but the mime library lists the mathml mime under "mathml" instead of the file extension "mml". probably a better way to do that but eh

fulfills #382

### DIFF
--- a/src/handlers/pandoc.ts
+++ b/src/handlers/pandoc.ts
@@ -241,13 +241,22 @@ class pandocHandler implements FormatHandler {
         [inputFile.name]: new Blob([inputFile.bytes as BlobPart])
       };
 
-      const { stderr } = await this.convert({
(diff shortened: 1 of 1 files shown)


p2r3:
> fulfills #382

In the future, consider using one of these keywords for linking an issue to a PR: https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests
(`Fulfills` is not one of those)

PR #418, opened 2026-02-23
add basic direct txt to mml conversion using <mtext>

bas1licate:
adding direct txt to mml conversion using the <mtext> element, which is used for text. incase anyone wants to do that.

### DIFF
--- a/src/handlers/ImageMagick.ts
+++ b/src/handlers/ImageMagick.ts
@@ -3,7 +3,8 @@ import {
   Magick,
   MagickFormat,
   MagickImageCollection,
-  MagickReadSettings
+  MagickReadSettings,
+  MagickGeometry
 } from "@imagemagick/magick-wasm";
 
 import mime from "mime";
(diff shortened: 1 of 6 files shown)


p2r3:
Your pull request fails to build:
> src/handlers/txt2mml.ts(38,33): error TS2339: Property 'text' does not exist on type 'FileData'.

Please test with `bun run build` before submitting a pull request for review.

bas1licate:
> Please test with `bun run build` before submitting a pull request for review.

how

bas1licate:
if it's covered by the workflows how do i get them to run

bas1licate:
actually i think i figured it out kinda

bas1licate:
ok i think i fixed it (slightly more specifically: bc1b3f5 checks: all successful except what doesn't matter)

bas1licate:
the check results for 6fd0570 only show up when looking at it in the fork so here's [a link to that](https://github.com/bas1licate/moremml/commit/6fd057073b30205d2f1ffaf2614e91d821dec653)

bas1licate:
@p2r3 it's been almost a week since i opened this pr can you look over it again soon

p2r3:
What does this do, exactly? Or rather, why is this necessary? Can't Pandoc already do this? A `txt -> mathml` conversion route already exists.

bas1licate:
oh ok

PR #421, opened 2026-02-23
Introduction of raw red, green, blue, alpha format

omega-fallon:
A companion to RGB. Only really works for square images.

### DIFF
--- a/src/handlers/rgba.ts
+++ b/src/handlers/rgba.ts
@@ -93,7 +93,7 @@ class rgbaHandler implements FormatHandler {
 
                 if (outputFormat.internal == "rgba") {
                     // Fill in with 255 in alpha channel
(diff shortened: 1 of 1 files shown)


src/handlers/rgba.ts
```diff
@@ -0,0 +1,196 @@
...
+                }
+            }
+            else if (inputFormat.internal == "rgb") {
+                if (new_file_bytes.length % 3 != 0) {
+                    throw new Error("Invalid RGB file size; not a whole number of samples.");
+                }
+
+                if (outputFormat.internal == "rgba") {
+                    // Fill in with 255 in alpha channel
+                    let writer_array = new Uint8Array(new_file_bytes.length * (1/0.75)); // this might cause floating point imprecision errors at super big file sizes
```
p2r3: Why not do this arithmetic more safely? You acknowledge the issue yet do nothing about it. A potential fix would be to avoid floating point math entirely (e.g. `length + Math.floor(length / 3)`), or even just wrapping it in `Math.round`.

Issue #423, opened 2026-02-23
[File format] Add support for High Efficiency Image File (HEIF / HEIC)


SPLATPLAYS:
**Describe the file format**
High Efficiency Image File (HEIF), commonly using the .heif or .heic extension, is a modern image container format based on HEVC (H.265). It supports high-compression still images, image sequences, transparency, and metadata.

It is widely used on Apple devices (iOS ≥ 11, macOS ≥ High Sierra) as the default photo format because it provides better compression than JPEG while maintaining similar or higher quality.

**Describe the expected behavior**
Users should be able to:

Convert HEIF/HEIC → PNG, JPEG, WebP, AVIF

Convert PNG/JPEG/WebP → HEIF/HEIC

When converting from HEIF, the primary image frame should be extracted and converted as a standard raster image. Metadata preservation would be a bonus but is not strictly required for initial support.

**Provide reference implementations**
Open-source implementations compatible with GPL-2.0:
libheif (Lesser GNU):
[https://github.com/strukturag/libheif](https://github.com/strukturag/libheif)

Widely used HEIF/HEIC decoder and encoder library.

heic2any (browser-based) (MIT License):
[https://github.com/alexcorvi/heic2any](https://github.com/alexcorvi/heic2any)

JavaScript-based HEIC converter using WebAssembly.

**Do you plan on implementing this yourself?**
Yes

**Additional context**
HEIF is increasingly common due to smartphone adoption, especially from Apple devices. Adding support would significantly improve compatibility for users uploading modern photos.

PR #430, opened 2026-02-23
Add HEIC/HEIF image format support #423

SPLATPLAYS:
#423


p2r3:
Where did you get this code?

SPLATPLAYS:
A lot of the decoding code is from the https://github.com/alexcorvi/heic2any docs. Encoding via native HTMLCanvasElement.toBlob is spec'd here https://html.spec.whatwg.org/multipage/canvas.html#dom-canvas-toblob-dev . The rest is glue code.

SPLATPLAYS:
@p2r3

p2r3:
I'm a bit suspicious of this pull request being vibecoded, along with possibly some of your others.

- The rate at which you're pumping these out, and how casual your pull requests seem to be, is unusual. It's not impossible by any means, but it also isn't normal for a single contributor to pick up such a wide range of formats, implement each in a single commit, and submit no description or additional notes.
- You've got some weird code comments. Some bother to point out the obvious, others don't. Some have Unicode characters in them (e.g `→`, `↔`, and `≥`), which would be annoying to type for a human. The style also seems overall inconsistent, as some are written in perfectly punctuated and capitalized English, while others are in all-lowercase with no punctuation.
- One of the error messages reads `Consider converting to PNG or JPEG instead`, which is incredibly bizarre for a file converter, _especially_ since the user isn't going to see this error. To me, this reads like an LLM slightly misunderstanding the greater context of the project it's working on.
- You deviate quite a bit from established conventions in the rest of the codebase. Your definition of `init` defines a return type when our examples and other handlers typically don't. You also just underutilize `init` in general - you define a canvas _during_ the conversion, whereas other handlers that require a canvas typically do it _before_.
- It also seems weird to me that you're referring to the HTML spec for `canvas.toBlob`, when that method has been used countless times throughout this codebase in some of the very earliest image handlers.

PR #439, opened 2026-02-24
Update infinite craft handler

ghost:
Add support for IC -> JSON and JSON -> IC


ghost:
prob should make sure the json file is actually an infinite craft save but idk how id do that yet

PR #442, opened 2026-02-24
Improved filetype detection for executables and added EXE -> BAT conversion

mdev34-lab:
- **Added** proper header-based filetype detection for executable binaries, allowing proper input filetype detection for Windows, Linux and Mac binaries rather than defaulting to `application/octet-stream`
- **Added** an `exe -> bat` conversion bridge, which encodes the Windows executable into a self-extracting batch file using base64, thus making the conversion between these filetypes bidirectional

Please inform me of any issues in the subitted commits. =)

### DIFF
--- a/src/handlers/index.ts
+++ b/src/handlers/index.ts
@@ -40,6 +40,7 @@ import txtToInfiniteCraftHandler from "./txtToInfiniteCraft.ts";
 import wadHandler from "./wad.ts";
 import espeakngHandler from "./espeakng.js"
 import textToShellHandler from "./texttoshell.ts";
+import exeToBatHandler from "./exeToBat.ts";
 import batchHandler from "./batch.ts";
 import bsorHandler from "./bsor.ts";
 import txtToPyHandler from './textToPy.ts';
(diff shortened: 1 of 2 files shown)


p2r3:
Firstly, please limit your pull requests to one feature per.

Secondly, if you were to implement MIME type detection from headers, it should likely be done with a library that maintains a wider database of formats, rather than implementing each formats by hand ourselves. I figure that in this case the idea was to support just executables to scaffold your own handler, but EXE and BAT files both have file extensions anyway. If you're going to implement this feature, please do it with foresight beyond your domain.

Thirdly, what's up with the massive diffs?? Every file you've touched is treated as having been entirely wiped and rewritten.

If this was split into two PRs, I would've likely merged the `exe->bat` handler and closed the other one, but now I'm not quite sure _what_ to do. I'll let you fix the diffs and revert the MIME detection if you want, or you can create a new PR.

mdev34-lab:
@p2r3 Done. Reverted the MIME detection shenanigans, now this PR only has the EXE -> BAT pipeline.

mdev34-lab:
Oh, and regarding the huge diffs, I guess they're probably due to line ending normalization (or the lack thereof) or something similar, since they show up normally in my VS Code instance.

src/handlers/index.ts
```diff
@@ -1,113 +1,115 @@
-import type { FormatHandler } from "../FormatHandler.ts";
```
exurd, line 1: Did you accidentally commit the file with CRLF line endings instead of LF?

mdev34-lab: Ohhhh that's why the diffs are messed up...

mdev34-lab: Done. Changed it to LF

exurd:
Can you please squash the commits so we don't have reverts or CR-LF fix commits when merging this PR?

mdev34-lab:
@exurd Done. Squashed the commits so everything looks clean

PR #443, opened 2026-02-24
Add CSV to SQLite to sqlite handler

forester302:


p2r3:
Not the prettiest code but it'll do.

In the future, consider moving those helper functions into the class, as well as maintaining consistent indentation.

PR #448, opened 2026-02-25
Improve image conversions to python

wenxigoh:
Edited TraversionGraph.ts to add a penalty to converting images to text then to python


p2r3:
Your pull request is empty?

exurd:
> Your pull request is empty?

Seems commit 96087dd caused this.

p2r3:
I'm actually not at all confident that this is the right solution.

> `// Image to code through text is very likely not what the user wants`

In most cases, image to code through text _is_ what the user wants. For anything other than Python, it's the only route. Punishing it so generally breaks all other `image -> code` paths that would normally go through ASCII art.

p2r3:
Closing as stale? Or, at least, I'm certainly not merging this as-is. If you want to work on a better solution, create a new pull request.

Issue #419, opened 2026-02-23
[Bug] Multiple file input not supported in file picker


GaspacoZanis:
https://github.com/user-attachments/assets/5ca4965a-79d0-4c15-b83c-ebd25b2f1cd6

In the attached video it can be seen that I can drag'n'drop more than 1 file of the same type, but I cannot do that after clicking the button for selecting the file. After clicking the button, the selection is limited to just 1 file even if I try to select multiple items. Is this difference in behaviour intentional or not?

I am using Windows, and I am on the latest version of Firefox.

PR #468, opened 2026-02-27
Fix issue #419: Allow multiple file selection in popup and improve selection of multiple file

PenguinCre8te:
Allow multiple file selection in popup and instead of showing "... and n more" actually list out the files

### DIFF
--- a/index.html
+++ b/index.html
@@ -9,7 +9,13 @@
 </head>
 <body>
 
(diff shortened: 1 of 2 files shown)


p2r3:
AFAIK this is one of the issues tackled by #73 - I'd rather just wait for that and not cause any more merge conflicts for now.

PR #470, opened 2026-02-27
fix: replace unstable jsr @std/yaml with npm yaml

ConnorTippets:
replaces the jsr-sourced yaml which has proven to not be a stable source, with the npm-sourced yaml


p2r3:
Huh, why wasn't this used before?

PR #473, opened 2026-02-28
feat: wld -> png conversion

### LINKED ISSUE #162: Support for .wld to .png

ConnorTippets:
just in case #244 doesn't end up working out, here's this
code basically ported from that pr (courtesy of pixelkat5), however now using a submodule'd library for wld parsing ([TEdit/terraria-world-file-ts](https://github.com/TEdit/terraria-world-file-ts/))

closes #162

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -21,4 +21,4 @@
 	url = https://github.com/ConnorTippets/RPG-Maker-MV-Decrypter.git
 [submodule "src/handlers/terraria-wld-parser"]
 	path = src/handlers/terraria-wld-parser
-	url = https://github.com/TEdit/terraria-world-file-ts.git
+	url = https://github.com/ConnorTippets/terraria-world-file-ts.git


ConnorTippets:
oh god what happened in the actions?!

it seems to be angry at the library

Celestemysterios:
does that mean theres something wrong with the library itself/it cant be used?

exurd:
~~Hmm... CtI's tsconfig.json says it's using `ES2022` while the library you're using is on `ES2024`...~~ Reading the files more, the target for CtI says `ES2022`, terraria-world-file-ts's target is `ES2020`.

Though if it's working on your end, it wouldn't make sense for this to be the reason why it's not working on actions...

https://github.com/p2r3/convert/blob/d11688ea4c2a8a0a713ed3160c0146840b3fafb9/tsconfig.json#L7
https://github.com/TEdit/terraria-world-file-ts/blob/594b0070c80d0b189fe037e333337ab85d434827/tsconfig.json#L8C8-L8C14

ConnorTippets:
> Though if it's working on your end,

it doesn't actually seem to be working on my end? deploying it with `bunx vite` works fine, but `bun run build` spits out the same errors that the action got

p2r3:
Pretty classic case of a dependency adhering to a less strict TS standard than us. You'll have to add some annotations to prevent the TypeScript compiler from checking those imports.

ConnorTippets:
got rid of some errors! still got the ones about es2024 and type-only imports, though

ConnorTippets:
looking back, "fix all build errors" is not a very descriptive commit.

Issue #453, opened 2026-02-26
[File format] request for .cbz and adjacent comic book file support


Shivyneko:
**Description**
This .cbz file format is just a zip file with images in it. It is used in comic book readers like komikku or mihon.

**Describe the expected behavior**
.pdf -> .png -> .zip -> .cbz or
Some image format -> .zip ->.cbz
Some audio format -> .png ->.zip ->.cbz

**Each page in the pdf should be extracted to png**
Pre-existing zip converter should be used to compress images into a .zip file and the extension should be changed (renamed) to .cbz , easy as that. It can contain .xml file for information of the chapter or series. **This can be fully ignored tho as it works fine even without it**.Example format from random beserk chapter:

\<?xml version=\'1.0\' encoding=\'UTF-8\' ?\\>
\<ComicInfo\> xmlns:xsd=\"http://www.w3.org/2001/XMLSchema\" xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\"\>
  \<Title\>Chapter 150\</Title\>
  \<Series\>Berserk\</Series\>
  \<Number\>150\</Number\>
  \<Summary\> summary or something\</Summary\>
 \<Writer\>Kentarou Miura, Studio Gaga\</Writer\>
  \<Genre\>Action, Drama, Super Power, Supernatural, Demons, Adventure, Fantasy, Horror, Military, Psychological, Seinen, Magic\</Genre\>
  \<Web\>website link for source or somthing\</Web\>
  \<ty:PublishingStatusTachiyomi xmlns:ty=\"http://www.w3.org/2001/XMLSchema\"\>Ongoing\</ty:PublishingStatusTachiyomi\>
  \<ty:Categories xmlns:ty=\"http://www.w3.org/2001/XMLSchema\"\> \</ty:Categories\>
  \<mh:SourceMihon xmlns:mh=\"http://www.w3.org/2001/XMLSchema\"\> \</mh:SourceMihon\>
\</ComicInfo\>

**Provide reference implementations**
Use the current .zip implementation only.

Poppler or some adjacent tool should be used for image extraction from pdfs.

**Justification** 
Pdf readers kinda suck at their whole job and this allows for an easier time.

**Do you plan on implementing this yourself?**
No

**Additional context**
Files should be first be converted to images if they are not already. .png should be the default for converting non images like audio.

Fjox:
youd want to add cbr too since that version is pretty popular too

omega-fallon:
[...]

PR #476, opened 2026-03-01
Support for converting an array of images to a CBZ file

omega-fallon:
This pull is related to, but does not entirely close issue #453. Any array of images gets renamed to pages and zipped into a renamed zip file. This handler also has support for zipping an array of images into a standard zip file, in which case no renaming is done.


omega-fallon:
As a quick addendum, I believe the mentioned issue is now actually closed with my changes.

PR #479, opened 2026-03-01
add tar support

neko782:
simple support for tar<>zip conversion using nanotar. coauthored with claude 4.6 opus, reviewed everything it wrote.


neko782:
changed some things i missed the first try

p2r3:
This really didn't need to be implanted with Claude. It's a simple library wrapper. Merging this would be a disrespect to actual human contributors.

PR #481, opened 2026-03-02
fixed lossless markers and categories

neko782:
some formats which obviously are lossless were not marked as such, and some other formats did not have a category - meaning they would always get the category cost (because it falls back to mime category and most formats have a non mime category)


neko782:
in conflict with #482, maybe merge that one first and then i can fix the merge conflicts

neko782:
i dont like this pr for some reason

PR #482, opened 2026-03-02
refactor: file format normalization

ConnorTippets:
between all the different handlers, many idiosyncrasies have arisen, such as many handlers being made before CommonFormats, and some handlers trying to be different with how they describe formats (looking at you imagemagick), and many duplicates
as such, this pr aims to remove as many of these duplicates, while also normalizing all file formats to 1 singular format and description (using CommonFormats if possible).

looking at the changes, i think src/handlers/petozip.ts had some sort of catastrophic line ending error


p2r3:
Loosely skimmed, nothing glaringly bad, I'll trust the tests. If this breaks something I'm making you eat a shoe 💔

PR #485, opened 2026-03-03
 fix txt and ordering issue with pandoc 

### LINKED ISSUE #434: [File format] RTF -> TXT
**Describe the file format**
RTF (Rich Text Format)

Expected behavior would be simply converting the document to plain text. Currently, this routes through JSON and produces garbage data.

I believe https://github.com/sigmaSd/stripRtf would be of use to this.

No

neko782:
fixes #434

there was a bug where we renamed pandoc's "plain" to "text" which caused it to not work at all and where json and xml routes would get picked instead of obviously better routes.

### DIFF
--- a/src/handlers/pandoc.ts
+++ b/src/handlers/pandoc.ts
@@ -175,7 +175,8 @@ class pandocHandler implements FormatHandler {
     outputFormats.forEach(format => allFormats.add(format));
 
     this.supportedFormats = [];
-    for (let format of allFormats) {
+    for (const internal of allFormats) {
+      let format = internal;
       // PDF doesn't seem to work, at least with this configuration
       if (format === "pdf") continue;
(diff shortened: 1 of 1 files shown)


src/handlers/pandoc.ts
```diff
@@ -175,7 +175,8 @@ class pandocHandler implements FormatHandler {
     outputFormats.forEach(format => allFormats.add(format));
 
     this.supportedFormats = [];
-    for (let format of allFormats) {
+    for (const internal of allFormats) {
+      let format = internal;
```
p2r3, line 179: Confused why this was necessary, but okay.

neko782: preserves the format pandoc gives in case we change the display name any other formats other than plain

PR #487, opened 2026-03-03
implemented tar handler (again)

neko782:
reimplemented tar but this time by hand. didnt look at the other pr while doing it, only remembered the general structure and that it used nanotar.


neko782:
sorry keep second guessing myself, fine now

p2r3:
I'll leave this open for a bit - let me know if you plan to implement the archive category thing I proposed in #480. I'm fine with merging either way.

neko782:
> I'll leave this open for a bit

the archive category thing isnt really blocking, the format already works today so i think we can merge. can implement it in a later pr (also i would have to change lzh handler code which i dont think fits here)

PR #492, opened 2026-03-04
Support for CBT

omega-fallon:
In combination with the new support for TAR archives, I've added simple support for the CBT (Comic Book TAR) format. This modifies my previously-added comics.ts handler as well as the new tar.ts handler.

### DIFF
--- a/src/handlers/comics.ts
+++ b/src/handlers/comics.ts
@@ -142,7 +142,7 @@ class comicsHandler implements FormatHandler {
             outputFiles.push({ bytes: bytes, name: baseName + "." + outputFormat.extension });
         }
         // Unpack a tar/cbt with code from tar.ts
-        else if ((inputFormat.internal === "tar" || inputFormat.internal === "cbt") && image_list.includes(outputFormat.internal)) {
+        else if ((inputFormat.internal === "cbt" || inputFormat.internal === "tar") && image_list.includes(outputFormat.internal)) {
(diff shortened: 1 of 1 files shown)


neko782:
oofie, i still dont like the idea of hardcoding every renamable format into every archive handler... (even though p2r3 was in support of it) isnt that what the rename handler is for? we just need a path going from the archive formats back to their renamed counterparts

neko782:
btw this now makes tarhandler claim that it can convert tar -> cbt or tar -> cbz and many of the other incorrect formats. also anyinput makes it claim that any file can be now converted to cbt or cbz but thats an issue with anyinput/tar.ts in general

but honestly i guess this is fine, i cant think of anything better other than "rename handler with validation" soo i guess this can stay

omega-fallon:
I'll admit I was a bit hesitant editing tar.ts in the first place (not my code) but it seemed better than just needlessly copying over two huge chunks of the handler, and it was necessary in order to make cbt<-> cbz possible, since you can rename a comic book archive and guaranteed get a valid archive, but not vice versa, the pathing wouldn't work out without that, unless I'm mistaken. Having the handler claim it can make conversions it can't might be bad, but as long as it throws an error instead of giving something malformed, is there an issue? The pathfinder would just skip it in that case, right?

neko782:
yeahhhh i guess this is the only way for now

PR #495, opened 2026-03-04
implement diffrent compressed tar formats

neko782:
only .tar.gz for now, planning to implement .tar.xz and .tar.zst


neko782:
didnt find a good library for compression so only decompression for now

neko782:
found one for zstd but not for xz

neko782:
ready to merge

PR #496, opened 2026-03-04
add llm use policy in readme

ConnorTippets:
adds a simple llm use policy subsection to the contributing section

writing from other people's POV and using their opinions is not my strong suit, so if I've said anything wrong or weird, please let me know.


neko782:
(lol why did build fail randomly)

i mean we do need a llm policy but that doesnt really say anything more than "you shouldnt make vibe coded prs sometimes". well i guess it does say that you shouldnt use an llm when it can be implemented by hand but is that enough?

p2r3:
I'll just add this with my own wording and reasoning, I feel like that'd make more sense.

PR #498, opened 2026-03-04
fix missing categories and flags on formats

neko782:
supersedes #481

adds missing stuff on the formats that handlers define

vibe coded because 1. i dont wanna do it 2. it needs to be done 3. its gonna do it much better than me


src/handlers/FFmpeg.ts
```diff
@@ -218,7 +218,9 @@ class FFmpegHandler implements FormatHandler {
       mime: "video/quicktime",
       from: true,
       to: true,
-      internal: "mov"
+      internal: "mov",
+      category: "audio",
```
neko782, line 222: if you are wondering thats .qta (QuickTime Audio) thats just mov

PR #499, opened 2026-03-04
Add WMV (Windows Media Video) support

StoneCotton:
Adds support for WMV input and output by mapping it to the ASF container
format already handled by FFmpeg.

**Changes:**
- Add WMV `FormatDefinition` in `CommonFormats.ts`
- Add `asf` → "Windows Media Video (WMV)" friendly name mapping in FFmpeg handler
- Add manual WMV format entry using `internal: "asf"` (same pattern as QTA)
- Add `video/x-ms-wmv` → `video/x-ms-asf` MIME normalization
- Set encode bitrate to 15 Mbps video / 192 kbps audio (FFmpeg's default ~200 kbps is too low for the older WMV codecs)

### DIFF
--- a/src/handlers/FFmpeg.ts
+++ b/src/handlers/FFmpeg.ts
@@ -19,7 +19,8 @@ class FFmpegHandler implements FormatHandler {
     ["matroska", "Matroska / WebM"],
     ["mov", "QuickTime / MOV"],
     ["3gp", "3GPP Multimedia Container"],
-    ["3g2", "3GPP2 Multimedia Container"]
+    ["3g2", "3GPP2 Multimedia Container"],
+    ["asf", "Windows Media Video (WMV)"]
   ]);
 
   public name: string = "FFmpeg";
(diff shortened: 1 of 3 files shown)


src/handlers/FFmpeg.ts
```diff
@@ -221,6 +222,18 @@ class FFmpegHandler implements FormatHandler {
...
+      name: "Windows Media Video",
+      format: "wmv",
+      extension: "wmv",
+      mime: "video/x-ms-asf",
+      from: true,
+      to: true,
+      internal: "asf",
+      category: "video"
+    });
+
```
ConnorTippets, line 236: ~~You added CommonFormats.WMV, why not use it?~~

Ah, I see what's happening here. Admittedly figuring this out myself was a bit difficult too, but FormatDefinition.builder takes a string that _is_ the internal. So CommonFormats.WMV.builder("asf").allowFrom().allowTo() would produce the same thing.

src/handlers/FFmpeg.ts
```diff
@@ -284,6 +297,8 @@ class FFmpegHandler implements FormatHandler {
       command.push("-vf", "setsar=1", "-target", "ntsc-dvd", "-pix_fmt", "rgb24");
     } else if (outputFormat.internal === "vcd") {
       command.push("-vf", "scale=352:288,setsar=1", "-target", "pal-vcd", "-pix_fmt", "rgb24");
+    } else if (outputFormat.internal === "asf") {
+      command.push("-b:v", "15M", "-b:a", "192k");
```
p2r3, line 301: I believe this is unnecessary. I'm leaving it in because I don't have the spare time to test if it works without it, but ideally we wouldn't be enforcing custom bitrates on outputs unless there's a good reason.

PR #505, opened 2026-03-06
Binary waterfall RGBs

omega-fallon:
Support for converting reading certain file types' (currently only .exe) bytes as raw color samples as a way of visualizing them (i.e. converting them to images) for cases where no other good visualization exists. Example:
RPG_RT


p2r3:
As I've said many times before, we shouldn't lean into dumping pure binary data. That defeats the point, see the readme. There _are_ ways to visualize an EXE's contents.

PR #521, opened 2026-03-08
feat: xcf -> png conversion

### LINKED ISSUE #452: [File format] XCF
**Describe the file format**
XCF is an image format specifically for GIMP, a free, open source, image editing program for Windows, Linux, and MacOS. Specifically, XCF stores the layers, text data, etc. that can each be separately manipulated in the program without losing any data. GIMP natively supports PNG, PDF, GIF, etc. but not files such as videos, text (as input), and exotic image formats.  GIMP can import PDF and SVG, but it rasterizes the content, losing any text data.

I'll just describe the more unorthodox formats that aren't really natively supported by GIMP.
XCF -> video: Each layer becomes its own frame, with the bottom-most layer becoming the first frame of the video. This process should be reversible.
PDF -> XCF: Each page becomes its own layer, possibly with the text, images, etc. becoming layers between the "page layers". The "page layers" should have the same dimension as the entire image, while the content layers should be smaller.
XCF -> PDF: Merge layers based on their dimensions, so that the "page layers" are merged with the correct content.
TXT -> XCF: Produce a file that contains an editable text layer with the same content as the text. This process should be reversible, but any images should be made into ASCII art.

Of course, XCF should still be able to be converted to and from image files.

I'm not a good searcher so I couldn't find any other people's implementations, but here's the official documentation
https://developer.gimp.org/core/standards/xcf/

No

**Additional context**
XCF appears to be a somewhat complex format, containing metadata that GIMP uses to verify if it is a valid file. Additionally, the format changes as GIMP updates, so there would either have to be separate choices for major versions or a single choice for a single format.  As said before, XCF is open source, so most of the tools have been already provided for you. 
 
From the official documentation:
"Authors of third-party XCF-creating software in particular should take care to write files that are **as indistinguishable as possible** from ones saved by GIMP. The GIMP developers take care to make each version of GIMP able to read XCF files produced by older GIMP versions, but they make no special efforts to allow reading of XCF files created by other software."

ConnorTippets:
i spent a week of my life writing a pure js gimp xcf save file parser for this ([ConnorTippets/gimper](https://www.github.com/ConnorTippets/gimper))

this doesn't fully close #452 but it is why i did this


p2r3:
holy moly

Supertrain12:
thx man

PR #525, opened 2026-03-09
Add APF (Aperture Picture Format) support

Coleisforrobot:
Added support for the 1-bit image format made by Valve for the Portal 2 release ARG. Everything is implemented directly in the handler, so no external libraries or anything. Only goes APF -> BMP and not the other way around, would require me write some encoder far above my level of expertise. A lot of it is based on the original BASIC program from Valve, which is at [the old ARG wiki](http://portalwiki.asshatter.org/index.php/Aperture_Image_Format.html). No colour support, as that relies on an external file (Aperture Menu Format) and I couldn't see an obvious way to get in a second file without requiring the user to zip it up or something; I left some information about that in the comments at the bottom if anyone wanted to go on that adventure.

I've been quite sick over the past few days and have mostly worked on this out of boredom/as a side-project so apologies if it isn't the cleanest, also apologies for no commits I worked on this mostly in Notepad++ w/ language highlighting.

The code works on my end (my apologies if there is an established way to spit out errors like to the user I couldn't find anything and the live one at convert.to.it seems to just hang if you give it malformed files so I guessed not), repeating myself in the comments but if anything doesn't work as expected lmk. As I mentioned I am quite sick atm so I might take a bit to respond.


ConnorTippets:
> would require me write some encoder far above my level of expertise.

not true! handlers already exist that convert to png, and it's completely possible without a library. look into the `canvas.toBlob` function.

ConnorTippets:
ah! I was under the impression you were asking about APF -> PNG or other similar image formats

src/handlers/aperturePicture.ts
```diff
@@ -0,0 +1,148 @@
...
+}
+
+function decodeAPF(data: string, SK: number): Uint8Array {
+  const w = 320,
+    h = 200;
+  if (SK <= 0) throw new Error("Malformed APF file (SK is invalid, <= 0)");
+  const bmp = new Uint8Array(w * h);
+  let x = 0,
+    y = h - 1,
+    draw = true, // no idea how this works. the original basic code from the ARG sets draw to false and then inverts it and when i implemented that it kinda worked but the colour was inverted so i removed the draw  = !draw and it refused to draw anything so i just flipped this to true and now it works. i'm way too ill for this
```
Coleisforrobot, line 66: note i mean as in idk how setting it to true works, the original sets it to false and then inverts but that doesn't produce the same results here. it might be because the original basic script expects colour meanwhile we have a forced black or white here, so i've interpreted it wrong. either way, the decoder works as expected now

PR #535, opened 2026-03-13
Add code format support (css, js, ts, cpp, java, php, rb, rs, go, lua, ino)

Oranged12:
All formats were added to the existing renameTxtHandler inside src/handlers/rename.ts, meaning no new files or dependencies were needed.

Since all of these formats are plain text under the hood, conversion works by simply renaming the file extension and routing through the existing plain text pipeline. This means they can be converted to and from any format that already supports .txt, such as .json, .xml and .yaml.

Both "from" and "to" are enabled for all formats, meaning you can convert INTO these formats as well as OUT of them. All conversions are marked as lossless since no data is lost when renaming a plain text file.

### DIFF
--- a/src/handlers/rename.ts
+++ b/src/handlers/rename.ts
@@ -1,6 +1,6 @@
 import CommonFormats, { Category } from "src/CommonFormats.ts";
 import type { FileData, FileFormat, FormatHandler } from "../FormatHandler.ts";
-
+ 
 // base class for handling renames
 function renameHandler(name: string, formats: FileFormat[]): FormatHandler {
   return {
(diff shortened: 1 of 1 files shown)


p2r3:
> Since all of these formats are plain text under the hood, conversion works by simply renaming the file extension and routing through the existing plain text pipeline.

Translating the underlying data _is not conversion_. Imagine if we treated SVG images as XML data because "they're just plain text under the hood". When a user converts a JS file to a C file, they expect to get a functional program. Yes, that's not an easy task, but nobody said it would be.

5cover:
Code is not plain text, but syntax trees. If you can somehow convert to and from syntax trees, it could allow converting between programming languages.
Conceptually: source language -> compile to IR (intermediate representation) -> reconstruct equivalent program in target language, preserving comments

neko782:
> If you can somehow convert to and from syntax trees, it could allow converting between programming languages.

that wouldnt really produce a functioning program tho, we need transpilers

PR #541, opened 2026-03-15
Add JSON5 conversion handler and document handler-level tests

5cover:
This PR adds support for converting between JSON and JSON5. It also formalizes a testing pattern for handlers with non-trivial logic by documenting optional handler-level unit tests and adding the first such test in the repository.

- added a dedicated `json5` handler using the `json5` package
- registered the handler in the main handler list
- normalized common JSON5 MIME aliases
- added a focused unit test at `test/handlers/json5.test.ts`
- updated the README contribution guidance to describe the handler test convention

## On testing

Before this PR, the repository already had project-level tests in `test/`, but there was no documented place for handler-specific unit tests. In practice, that left an underspecified gap for handlers that do meaningful parsing, serialization, or filename logic that is hard to validate cleanly through traversal-only tests.

This PR makes the policy explicit:

- broad project-level behavior tests remain in `test/`
- optional handler-specific unit tests live in `test/handlers/<handlerName>.test.ts`

The intent is not to require a dedicated test file for every handler. The new convention is meant for handlers with meaningful internal logic where focused unit coverage is warranted.

JSON5 is the first handler to use this pattern because it directly parses and serializes structured text with syntax that differs from strict JSON.

## Unit Test Coverage

The new handler test covers:

- parsing JSON5 input with comments, unquoted keys, and trailing commas
- converting JSON5 input into strict JSON output
- converting JSON input into JSON5 output
- round-tripping JSON -> JSON5 -> JSON to verify that the handler preserves the underlying data correctly

## Testing

- `bun test test/handlers/json5.test.ts`

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,7 +156,7 @@ This project currently uses two levels of tests:
 - Broad project-level tests live directly in `test/` (for example graph traversal and end-to-end conversion smoke tests).
 - Optional handler-specific unit tests live in `test/handlers/`, using the file name pattern `<handlerName>.test.ts`. These are a good fit for handlers with meaningful parsing, serialization, or file-naming logic that is hard to exercise reliably through traversal alone.
 
-Not every handler needs a dedicated unit test, but handlers with non-trivial internal logic generally should have one.
+Not every handler needs a dedicated unit test, but handlers with non-trivial custom internal logic may benefit from having one.
 
 ### Adding dependencies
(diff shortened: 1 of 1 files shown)


p2r3:
In the future, please consider making separate pull requests for separate tasks (e.g. JSON5 support and handler unit tests).

PR #544, opened 2026-03-15
implement full xz support using liblzma

neko782:
(was holding off on this one until targz was merged and this needs to be seperate because its not as straight forward) 

implemented full xz support using bindings to liblzma as there are literally *zero* browser js libraries for compressing xz on npm or anywhere. chetgtp helped me with making it build reproducebly (https://github.com/neko782/xz-wasm) and the various weird (but apparently standard) ways to put bytes into memory in js, otherwise wrote everything myself

i took a bit of a shortcut in decompressing, instead of implementing full streaming with xz (which would be hell) i just try to decompress with bigger output buffers (usually it gets it in 1-3 iterations unless its something very compressable). lmk if thats an issue


src/handlers/tarCompressed/xz.ts
```diff
@@ -0,0 +1,151 @@
...
+    e.free(outPosBuf);
+  }
+}
+
+export function decompress(input: Uint8Array): Uint8Array {
+  const e = instance.exports;
+
+  let outputLength = input.length * 3; // we cant really predict output length so we just guess
+
+  for (let i = 0; i < 50; i++) { // limit just in case but we will probably hit memory before that
```
p2r3: Wasm has a memory limit of 2 GB, should probably gate somewhere before that instead? I fear that simply letting it run out of memory might cause the tab to hang.

neko782: > might cause the tab to hang

from all my testing it just crashes the handler after trying diffrent sizes for some time, ill make a better limit tho real quick

PR #545, opened 2026-03-15
consolidate source code converters into one

neko782:
mostly kept the same behavior, improved shell, also added rust, js, c and c++. 

json.stringify isnt a perfect match for some languages, but it is good enough that 99% of anything you throw at it will work.


p2r3:
honestly goated pr, the spread of cheap "txt to code" handlers was starting to mess with me

Issue #540, opened 2026-03-15
[File format] .Piskel files


BeansBeefBroccoli:
**Describe the file format**
.piskel files are a json-based pixel art files. They come from [Piskel](https://www.piskelapp.com).

**Describe the expected behavior**
As on the website, they should be able to be converted into animated GIFs, PNGs, ZIP archives containing one PNG for each frame, and C files. Below are screenshots showing export options in Piskel:
<img width="312" height="191" alt="Image" src="https://github.com/user-attachments/assets/bf84ce3c-7d85-4a55-a08a-86c4ed26aace" />
<img width="300" height="458" alt="Image" src="https://github.com/user-attachments/assets/b16a1b2c-2e6b-4377-936f-295dc45b2d5a" />
<img width="311" height="237" alt="Image" src="https://github.com/user-attachments/assets/8007053a-10cb-4341-9bd8-94edbf20d820" />
<img width="308" height="190" alt="Image" src="https://github.com/user-attachments/assets/801d8358-aa71-4006-a3f0-fd5592313124" />

**Provide reference implementations**
https://www.piskelapp.com/

**Do you plan on implementing this yourself?**
No

**Additional context**
[Some example Piskel files](https://gist.github.com/BeansBeefBroccoli/1614f3f0168599465e6e4c6b944681d8)

ConnorTippets:
idk if linking the app counts as reference implementation for the file format but sure

PR #554, opened 2026-03-17
feat: piskel -> spritesheet/zip

ConnorTippets:
`.piskel`s are sprites saved by https://www.piskelapp.com/

`piskel` -> `png`: exports a spritesheet (note, all layers are combined into one sprite per frame. idk if this is good design)
`piskel` -> `zip`: exports a zip with all of the frames (again, all layers are combined into one sprite per frame)

planned:
`piskel` -> `gif`: gif of frames that respects the piskel's defined framerate (object.piskel.frameRate once parsed)

doesn't fully close, but does address #540

### DIFF
--- a/index.html
+++ b/index.html
@@ -6,6 +6,16 @@
   <link rel="stylesheet" href="style.css">
   <link rel="shortcut icon" href="favicon.ico" type="image/x-icon">
   <title>Convert to it!</title>
+  
+  <meta name="url" content="https://convert.to.it">
+  <meta name="identifier-URL" content="https://convert.to.it">
+  <meta name="og:url" content="https://convert.to.it"/>
+  <meta name="og:title" content="Convert to it!"/>
+  <meta name="og:site_name" content="Convert to it!"/>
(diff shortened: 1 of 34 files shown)


p2r3:
I think this is probably sufficient to close #540. It's niche enough that I doubt it'll warrant revisiting.

src/handlers/piskel.ts
```diff
@@ -0,0 +1,168 @@
...
+                name: "Piskel Sprite Save File",
+                format: "piskel",
+                extension: "piskel",
+                mime: "image/png+json",
+                from: true,
+                to: true,
+                category: "image",
+                internal: "piskel",
+                lossless: true
+            }
```
p2r3, line 34: Why are these all marked as supporting to/from if the handler seems to only support Piskel input?

ConnorTippets: ah shoot, i forgot to remove that. i was originally planning to support the inverse (png/zip -> piskel) but decided not to.

PR #556, opened 2026-03-17
Fix OBJ format name

BunnyNabbit:
Corrects the file format's name defined in the threejs handler to Wavefront OBJ. This makes it consistent with the format name introduced by #555.


p2r3:
Wow that's an embarrassing typo, thanks

PR #562, opened 2026-03-18
feat: config support

Pendonym:
Add support for all the main configuration file formats

### DIFF
--- a/src/handlers/config.ts
+++ b/src/handlers/config.ts
@@ -62,6 +62,16 @@ class configHandler implements FormatHandler {
         internal: "ini",
         lossless: true,
       },
+      {
+        name: "RIVALS Configuration File", // https://www.roblox.com/games/17625359962
+        format: "txt",
+        extension: "txt",
+        mime: "text/plain",
+        from: true,
+        to: true,
+        internal: "txt",
+        lossless: true,
+      },
     ];
     this.ready = true;
   }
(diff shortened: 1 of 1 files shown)


src/handlers/config.ts
```diff
@@ -0,0 +1,120 @@
...
+class configHandler implements FormatHandler {
+  public name: string = "cfg";
+  public supportedFormats?: FileFormat[];
+  public ready: boolean = false;
+
+  private readonly cfgExt = ['ini', 'cfg', 'cnf', 'conf', 'cf'];
+
+  async init() {
+    this.supportedFormats = [
+      CommonFormats.JSON.builder("json").markLossless().allowFrom(true).allowTo(true),
```
p2r3: I don't think `lossless` applies here - JSON has some type information that plaintext configs don't.

src/handlers/config.ts
```diff
@@ -0,0 +1,120 @@
...
+          bytes: encoder.encode(stringified),
+        });
+      }
+    }
+
+    if (outputFiles.length === 0) {
+      throw new Error(
+        `configHandler does not support route: ${inputFormat.internal} -> ${outputFormat.internal}`,
+      );
+    }
```
p2r3, line 166: Wouldn't it make sense to support config->config by simply renaming the file? I get that this handler assumes all text-based configs are `ini` anyways.

src/handlers/config.ts
```diff
@@ -0,0 +1,120 @@
+import type { FileData, FileFormat, FormatHandler } from "../FormatHandler.ts";
+import CommonFormats from "src/CommonFormats.ts";
+import * as ini from 'ini';
+import * as yaml from 'yaml';
```
p2r3: Unused yaml import?

src/handlers/config.ts
```diff
@@ -0,0 +1,120 @@
...
+
+class configHandler implements FormatHandler {
+  public name: string = "cfg";
```
p2r3: Handler name should be the same as the file and class name. In this case, `config`, not `cfg`.

src/handlers/config.ts
```diff
@@ -0,0 +1,120 @@
...
+    inputFiles: FileData[],
+    inputFormat: FileFormat,
+    outputFormat: FileFormat,
+  ): Promise<FileData[]> {
+    const outputFiles: FileData[] = [];
+    const decoder = new TextDecoder();
+    const encoder = new TextEncoder();
+
+    const cfgRegex = new RegExp(`\\.(${this.cfgExt.join('|')})$`, 'i');
+    const yamlRegex = new RegExp(`\\.(${this.yamlExt.join('|')})$`, 'i');
```
p2r3: Unused yaml handling code

src/handlers/config.ts
```diff
@@ -0,0 +1,172 @@
...
+      {
+        name: "RIVALS Configuration File", // https://www.roblox.com/games/17625359962
+        format: "txt",
+        extension: "txt",
+        mime: "text/plain",
+        from: true,
+        to: true,
+        internal: "txt",
+        lossless: true,
+      },
```
p2r3, line 74: Where did this come from??

A text configuration for a random Roblox game is incredibly out of scope for a general-purpose "config" handler, and for this pull request. There's also no distinction between this and a regular text file, given that all of `format`, `extension`, and `mime` are identical to a plaintext file.

p2r3:
Superseded by #576

PR #564, opened 2026-03-18
Add `wabtHandler`

akiramusic000:
This PR adds a handler for `wabt.js`, which can convert `.wasm` files to `.wat` files, and vice versa.


src/handlers/wabtHandler.ts
```diff
@@ -0,0 +1,89 @@
+import type { FileData, FileFormat, FormatHandler } from "../FormatHandler.ts";
```
p2r3, line 1: Handler files should follow the naming convention `<name>Handler.ts`

akiramusic000:
Ah, thanks. Just fixed it.

src/handlers/wabtHandler.ts
```diff
@@ -0,0 +1,89 @@
...
+import { Category } from "src/CommonFormats.ts";
+import wabt from "wabt";
+
+const wabtModule = await wabt();
```
p2r3: Please avoid top-level awaits, this should be in `init`. The whole reason `init` exists is to let us lazy-load stuff like this.

You'll probably also have to add the functions below as methods on the handler's class.

akiramusic000: Done!

PR #567, opened 2026-03-20
add shToElf handler

neko782:
a pretty crude method but works and im proud of it


ConnorTippets:
the xkcd reference is Krazy

neko782:
a number had to be chosen (with a very secure rng) :p

src/handlers/shToElf/stub.c
```diff
@@ -0,0 +1,21 @@
+// gcc -o stub.elf stub.c
+
+#include <fcntl.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+
+// https://xkcd.com/221
+int cmd_size = 1273991571;
```
p2r3, line 9: Is there really no reason behind this number? Could this at least be a product of powers of two or something? Or really just some clean-ish number, or something chosen through some reason, rather than literally random?

I propose `1073741824` (or 1024^3), it's exactly 1 GiB and close to this value.

neko782: > Is there really no reason behind this number? Could this at least be a product of powers of two or something? Or really just some clean-ish number, or something chosen through some reason, rather than literally random?

the number is patched out by the js code with the real cmd size (important that its exact! we need to seek to the place in the binary where the shell script is located), i had to pick a number that i know wouldnt appear in the binary

p2r3: Oh that's... disgusting. Sure!

neko782: its super gross

src/handlers/shToElf/stub.c
```diff
@@ -0,0 +1,21 @@
...
+int main() {
+  int fd = open("/proc/self/exe", O_RDONLY);
+  lseek(fd, -cmd_size, SEEK_END);
+  
+  char *cmd = malloc(cmd_size + 1);
+  read(fd, cmd, cmd_size);
+  close(fd);
+
+  cmd[cmd_size] = '\0';
+  execl("/bin/bash", "bash", "-c", cmd, NULL);
```
Habetdin, line 20: So it heavily relies on bash and ignores other possible shells?

neko782: > So it heavily relies on bash and ignores other possible shells?

yes

cyklusiguess: i'm crying

Issue #569, opened 2026-03-21
[Bug] Converting multiple images through ImageMagik gives singular output


omega-fallon:
**Describe the bug**
Converting multiple images at once through ImageMagik only gives a single output, the first file.

**To Reproduce**
HEIC -> JPEG/PNG/etc. or any other image-to-image through ImageMagik, manually specified in Advanced Mode.

**Expected behavior**
All images get converted, not just one.

shanliuling:
I'll work on this. Could you please assign it to me?

p2r3:
Despite two pull requests, this still needs work. Potential implementers should see the note on #570.

PR #570, opened 2026-03-22
ImageMagik multi-input fix (closes issue #569)

omega-fallon:
Allows ImageMagik to properly process multiple InputFiles.

### DIFF
--- a/src/handlers/ImageMagick.ts
+++ b/src/handlers/ImageMagick.ts
@@ -85,6 +85,7 @@ class ImageMagickHandler implements FormatHandler {
     inputFormat: FileFormat,
     outputFormat: FileFormat
   ): Promise<FileData[]> {
+    const outputFiles: FileData[] = [];
 
     const inputMagickFormat = inputFormat.internal as MagickFormat;
     const outputMagickFormat = outputFormat.internal as MagickFormat;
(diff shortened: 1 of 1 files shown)


p2r3:
In some ways, this is a regression.

Currently, containers that may contain multiple images (e.g. PDF or GIF) combine their inputs to create multiple pages/frames. That's why the system is the way it is. Your fix breaks that.

A more correct solution would be to check that the output container supports multiple images, and adjust the behavior accordingly. I'm not entirely sure how you'd do this automatically, though. Hard-coding formats would be dirty.

omega-fallon:
> In some ways, this is a regression.
> 
> Currently, containers that may contain multiple images (e.g. PDF or GIF) combine their inputs to create multiple pages/frames. That's why the system is the way it is. Your fix breaks that.
> 
> A more correct solution would be to check that the output container supports multiple images, and adjust the behavior accordingly. I'm not entirely sure how you'd do this automatically, though. Hard-coding formats would be dirty.

I've put together a dirty fix on this branch. I also have no idea how this would be done automatically, but at least now it gives the desired behavior for both types of conversions.

omega-fallon:
Could I have a re-review on this if you get the chance @p2r3? Let me know if there's any formats I missed or if you have any thoughts on a non-dirty solution.

PR #571, opened 2026-03-22
fix: convert multiple images individually instead of merging (#569)

### LINKED ISSUE #569: [Bug] Converting multiple images through ImageMagik gives singular output

shanliuling:
## What is the problem?

Converting multiple images at once through ImageMagick only gives a single output (the first file).

## How did I fix it?

Changed the conversion logic to process each input file individually in a loop, rather than merging all images into one output.

- Loop through each inputFile in inputFiles array
- Convert each file separately
- Push each result to the 
esults array
- Return all converted files

## How was this tested?

- Syntax check passed
- Follows existing code patterns
- Logic verified against the bug description

## Usage

Users can now convert multiple images (e.g., HEIC -> JPEG) and receive individual output files for each input.

Closes #569

### DIFF
--- a/src/handlers/ImageMagick.ts
+++ b/src/handlers/ImageMagick.ts
@@ -92,22 +92,22 @@ class ImageMagickHandler implements FormatHandler {
     const inputSettings = new MagickReadSettings();
     inputSettings.format = inputMagickFormat;
 
+    const results: FileData[] = [];
 
-    const bytes: Uint8Array = await new Promise(resolve => {
-      MagickImageCollection.use(outputCollection => {
-        for (const inputFile of inputFiles) {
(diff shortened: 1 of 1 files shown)


omega-fallon:
Oh dang, looks like we had the same idea... I guess it's up to @p2r3 which pull request gets merged in. Frankly I don't care much as long as the bug is fixed.

p2r3:
See #570

PR #574, opened 2026-03-23
Improved handling of comic book archive and other archive formats

omega-fallon:
Various changes made for the sake of simpler and more expected conversion between archive formats.


src/TraversionGraph.ts
```diff
@@ -62,11 +62,18 @@ export class TraversionGraph {
         {from: "image", to: "text", cost: 0.5}, // Depends on the content and method, but can be relatively efficient for simple images
         {from: "text", to: "audio", cost: 0.6}, // Somewhat lossy for anything that isn't speakable text
         {from: "document", to: "text", cost: 1}, // Often very lossy, loses rich formatting
+        {from: "image", to: "image archive", cost: -1}, // Prioritize immensely
+        {from: "image archive", to: "image archive", cost: -2}, // Prioritize more
```
p2r3: Firstly, not a fan of negative weights. There's almost certainly a better way to accomplish whatever this is doing.

Secondly, can you explain _why_ you had to do this? You really shouldn't mess with weights unless absolutely necessary. I suspect that this is "patching" some problem, rather than fixing the root cause.

omega-fallon: First line is because the sevenZip converter has "any input" enabled, which causes slow routing if the input is some non-image and the output is an image archive. While I could split the converter in two such that there's a separate, image input–only converter for compiling image archives, that'd be a lot of duplicated code. The second line looking back is probably redundant with lines 72-75, so I'll remove it.

omega-fallon:
This should (🤞) be all set to go now.

neko782:
can we not avoid the new category?... feels a bit bad to introduce it just for this

neko782:
whar

neko782:
anyway, @omega-fallon i couldnt replicate the thing you were having that caused you to add these categories, so i just threw them out. please gimme some examples cuz i tested like png>cbz zip>cb7 cbz>tar cbz>cbt i didnt get anything bad enough

neko782 (approved):
hope you make more prs at some point

PR #576, opened 2026-03-23
Add support for JSON5,JSONC,YML,TOML,INI to and from JSON

Explosion-Scratch:
I just called this config.ts but I'm aware that #562 also is working on something similar, maybe I should rename to somthing else but I'm unsure what as these file formats are mainly used for configuration.


p2r3:
- If you see that someone is already working on something similar to what you're working on, don't just remake the same thing from scratch?? If the other pull request gets merged, yours will just get closed due to being mostly a duplicate. If _your_ pull request gets merged instead, then that's disrespectful to the author of the first one. I get that you've added more stuff in general, but you could've just waited a bit for the first PR to merge and worked on top of that.
- Your CSS pull request is on the same branch as this one, so you've dragged the commit from that into this. This not only makes the CSS PR redundant, it also bleeds the scope of this one.

Lucky for you, it seems #562 has made some strange choices, so I do feel like I'd rather merge your implementation instead. I also don't see any issues with your CSS handler, so I'm fine with merging that here. This 🤏 close to closing, though.

PR #580, opened 2026-03-24
Go between from and to on mobile

Explosion-Scratch:
This PR adds a back button on mobile, so essentially the flow on mobile looks like:
1. Upload file (same as before)
2. Choose input format for said file
3. The "Convert" button shows "Next", clicking "Next" confirms the input format and switches to output format selection
4. User chooses output format for file conversion
5. Button changes to say "Convert" and a back button is also added to re-select the input format.

https://github.com/user-attachments/assets/97dcde75-776c-4343-b80d-f43701b90dc2

Note: Should I create a new utils file for the new mobile related functions? I just put them at the top of main.ts but I can compartmentalize a bit if needed.

### DIFF
--- a/style.less
+++ b/style.less
@@ -178,7 +178,6 @@ button.selected {
   padding: .5em 1em;
   box-shadow: 0 4px 6px -1px rgb(0 0 0 / 0.1), 0 2px 4px -2px rgb(0 0 0 / 0.1);
   border: 2px solid #0003;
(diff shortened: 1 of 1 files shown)


Explosion-Scratch:
I've restyled things a tad, trying to improve functionality and usability but keep the same feel as before, here's what it looks like now:

image

image

p2r3:
Significant UI-related changes are currently being held in favor of #73.

PR #581, opened 2026-03-24
Draft: Add progress support

Explosion-Scratch:
Draft for adding progress support and abort signal, handlers get passed in doConvert. New param for doConvert is ConvertContext:
```js
/** Context passed to doConvert for progress/log reporting. */
export interface ConvertContext {
  /**
   * Report progress.
   * @param message - Status message to display.
   * @param value - Absolute 0–1 value, or updater `(old) => new`.
   */
  progress: (message: string, value: number | ((prev: number) => number)) => void;
  /**
   * Append a timestamped log entry.
   * @param message - The log message.
   * @param level - Log level ('log', 'error', 'debug', 'warn'). Defaults to 'log'.
   */
  log: (message: string, level?: LogLevel) => void;
  /** AbortSignal for cooperative cancellation. */
  signal: AbortSignal;
  /** Throws if the signal has been aborted. Call between async steps. */
  throwIfAborted: () => void;
}
```

This allows in doConvert to for instance at convenient times `ctx.throwIfAborted()` or `ctx.progress("Transcoding...", 0.5)`

### DIFF
--- a/index.html
+++ b/index.html
@@ -3,7 +3,7 @@
 <head>
   <meta charset="UTF-8">
   <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0">
-  <link rel="stylesheet" href="style.css">
+  <link rel="stylesheet" href="style.less">
   <link rel="shortcut icon" href="favicon.ico" type="image/x-icon">
   <title>Convert to it!</title>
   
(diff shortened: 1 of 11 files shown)


Explosion-Scratch:
Note I've continued from #580

p2r3:
Closing, see #580

Explosion-Scratch:
But what about the progress functionality only as opposed to the UI tweaks - this is a lot of under the hood stuff.

Explosion-Scratch:
This allows progress and logs via the doConvert method which now exposes a `ctx` param - I think this is a good approach regardless of the frontend but once #73  is merged (if it is) then I will re-PR with the backend changes to integrate with that frontend.

PR #583, opened 2026-03-25
EPUB to PDF improvement - EPUB -> HTML -> Typst -> PDF

Explosion-Scratch:
Added epub.ts which converts EPUB to HTML using epubjs. Extends Pandoc for HTML -> Typst (allowing asset inlining), registers epubjs-html pandoc and typst earlier so that path prefers them.

Result is correct EPUB to PDF conversion with pagebreaks and formatting, and EPUB to HTML looks amazing.


neko782 (approved):
good enough

PR #584, opened 2026-03-25
PPTX -> PDF

Explosion-Scratch:
Add PPTX to PDF by first converting to HTML then HTML2Canvas'ing it - in the future converting to SVG then SVG via Typst would be optimal


test/commonFormats.test.ts
ren-lv, line 1: This test file is not meant for checking whether specific handlers are used

test/commonFormats.test.ts
```diff
@@ -158,23 +174,71 @@ test("mp3 → png → gif", async () => {
...
+      "application/pdf",
+    ],
+  ].some(
+    (expected) =>
+      expected.length === conversion!.path.length
+      && expected.every((m, i) => conversion!.path[i].format.mime === m),
+  )).toBe(true);
+  expect(Object.values(conversion!.files[0].bytes).length).toBeGreaterThan(1000);
+
+}, { timeout: 60000 });
```
ren-lv, line 206: Why has this test been made more permissive?

PR #589, opened 2026-03-26
Better HTML to SVG, and lossless HTML to PDF via Typst

Explosion-Scratch:
The HTML to SVG handler now uses [this library](https://github.com/felixfbecker/dom-to-svg) to convert DOM to SVG (which works impressively well), then SVGs can be embedded into Typst to create vector PDFs!


Explosion-Scratch:
Here is what the demo word document turns into!

[word (3).pdf](https://github.com/user-attachments/files/26279389/word.3.pdf)

Explosion-Scratch:
This can also greatly improve #584 because it can then render to SVG -> PDF via this process.

Note: If #583 is merged then the process of EPUB directly to PDF should be preserved as EPUB -> HTML -> Typst (pandoc) -> PDF works much better in that case

Explosion-Scratch:
I'm not sure why tests are failing - they work locally. Maybe a cache thing?

ren-lv:
I'm not really gonna merge this with the failing test :/

neko782:
it is, in fact, failing

Issue #559, opened 2026-03-17
[File format] .c -> .txt


akiramusic000:
**Describe the expected behavior**
Currently, when converting `.c` files to `.txt` files, they are first converted to `.json`, and then to `.txt`. This is not ideal, as converting directly from `.c` to `.txt` by just changing the extension is easier, and more lossless.
This can also be extended to work with other programming languages.

**Provide reference implementations**
N/A

**Do you plan on implementing this yourself?**
Yes

**Additional context**
I'm not exactly sure how this should be implemented, as there is a conversion for `.txt` to `.c` (and other programming languages) that creates source code to print the text. Because of this, converting from `.c` to `.txt` to `.c` would be lossy, as instead of doing nothing, it would create source code that would print the original source code.

akiramusic000:
Alternatively, another option would be to somehow run the program and convert to it's output? This may be difficult, but it's probably possible since `clang` can be compiled to WASM.
EDIT: This would be lossy, however.

neko782:
i could see this working because sometimes you do want source files to be treated like text (like when converting to an image or just to txt itself), its not really unexpected and a valid source file *is* a valid text file

also related to #535

p2r3:
From the readme:
> When implementing/suggesting a new file format, please treat the file as the media that it represents, not the data that it contains. For example, if you were making an SVG handler, you should treat the file as an image, not as XML. In other words, avoid simple "binary waterfalls", as they're not semantically meaningful.

and from the issue @neko782 linked (#535):
> Translating the underlying data is not conversion. Imagine if we treated SVG images as XML data because "they're just plain text under the hood". When a user converts a JS file to a C file, they expect to get a functional program. Yes, that's not an easy task, but nobody said it would be.

What you're describing _is not conversion_. What's the point of simply renaming a C file to a text file? What value does that provide? Sure, it'd be less "lossy", the same way that converting an SVG to PNG would be less lossy if you just rendered the XML as black text on a white background. With OCR, it could even be lossless! But what value would that provide to the user?

C files contain code for a program. That's their medium, they should be treated in respect to that. We do have lots of file renaming handlers, but they're crutches, not valid solutions. So yes, as you later suggested, we'd ideally have a handler that compiles or executes the C file and prints the output of that. It's not the easy way out, but it's the only solution that makes sense.

neko782:
i mean renaming a c -> txt makes just as much sense to me as renaming a docx -> zip or a jar -> zip

[...]

PR #594, opened 2026-03-29
Added .sh files to the allowed text based formats

NilsBriggen:
Very small change, but without it .sh files are almost completely unsupported (or very cumbersome), even though they are basically text files.


ConnorTippets:
hello?? why did copilot review this??

neko782:
see #559 (even tho i disagre)

PR #604, opened 2026-04-01
add "alternate" mime support for message/coffeepot

bas1licate:
in the rfc `message/coffeepot` is specified in, there is a known erratum which accidentally mentions the type `application/coffee-pot-command`. since the erratum occurs earlier than the first actual mention of `message/coffeepot`, people are likely to use it accidentally. so just in case that happens, i added a case to `NormalizeMimeType.ts`


bas1licate:
now don't get confused. this pr is as serious as the rfc it stems from. i'm not doing a weird joke here.

ren-lv:
I'm admittedly still a bit confused, I didn't find mentions of coffeepot anywhere?

PR #610, opened 2026-04-03
fix: resolve infinite loop in path search algorithm

### LINKED ISSUE #605: [Bug] PPTX to PDF
Trying to convert office pptx file to pdf, finding conversion route takes too long and never finishes, creates a infinite loop I think because I've seen the same route couple of times before I cancel the search. The reason I tried to convert pptx to pdf is I didnt have office app installed, i know powerpoint have built-in converter for pdf. Love your work, thanks for the app.

**To Reproduce**
1. Uploaded the pptx
2. Selected the pptx format on the left menu
3. Selected the pdf format on the right menu
4. Pressed convert

**Expected behavior**
It should give a pdf file of the pptx, 

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Firefox
 - Version Firefox 149.0

**Additional context**
I know this task is simple and there is no need to use the website for it but I think it should convert it. I uploaded the file I tried to convert, it's not confidential, it's just lecture notes in turkish

[Isletme_Yonetimi__1-8_HAFTA_.pptx](https://github.com/user-attachments/files/26438486/Isletme_Yonetimi__1-8_HAFTA_.pptx)

.

Waleed-Ahmad-dev:
## What does this PR do?
This PR fixes a bug where the graph traversal algorithm gets caught in an infinite loop during complex format conversions (like `pptx` to `pdf` via pandoc). 

Fixes #605 

## Why is it needed?
Because `pandoc` registers a massive number of input/output combinations (creating over 4,000 edges), the pathfinding algorithm was experiencing an exponential path explosion. The custom `visitedBorder` logic was failing to properly prune redundant paths, causing the browser/app to hang indefinitely.

## How was it fixed?
* Replaced the global `visited` array and `visitedBorder` logic with a `minCosts` Map.
* Implemented standard Dijkstra pruning: the algorithm now records the lowest cost to reach any specific format node.
* If a newly discovered path to a node costs more than or equal to a previously found path, it immediately discards it, preventing infinite loops and drastically speeding up the search.
* Removed the now-obsolete `visitedBorder` property from the `QueueNode` interface.

## How was it tested?
* Tested locally by attempting a conversion from `pptx` to `pdf`.
* Verified that the algorithm now resolves the route almost instantly without hanging or crashing the app.

### DIFF
--- a/src/TraversionGraph.ts
+++ b/src/TraversionGraph.ts
@@ -5,7 +5,6 @@ interface QueueNode {
     index: number;
     cost: number;
     path: ConvertPathNode[];
-    visitedBorder: number;
 };
 interface CategoryChangeCost {
     from: string;
@@ -18,7 +17,6 @@ interface CategoryAdaptiveCost {
     categories: string[]; // List of sequential categories
     cost: number; // Cost to apply when a conversion involves all of the specified categories in sequence.
 }
-
 
 // Parameters for pathfinding algorithm.
 const DEPTH_COST: number = 1; // Base cost for each conversion step. Higher values will make the algorithm prefer shorter paths more strongly.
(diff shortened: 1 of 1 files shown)


5cover:
I was about to open an issue on  this very bug. Good catch.

ren-lv:
Seemingly doesn't solve the issue? In my testing this issue can only be reproduced if you have the cache -

bun run build
bun run cache:build
then move the cache from the dist directory into the project directory

Issue happens both in the current version and this PR. No notable speedups have been observed without the cache.
Feel free to reopen if I missed something.

PR #622, opened 2026-04-14
Error improvement

omega-fallon:
Edits thrown errors across the board to be more specific and transparent. Changes most handlers.


src/handlers/libopenmpt/libopenmpt.js
ren-lv, line 1: These seem unchanged and identical?

PR #624, opened 2026-04-16
Add Blockbench bbmodel to OBJ conversion support

valipuha:
## Summary

This change adds initial support for Blockbench `.bbmodel` files by introducing a focused `bbmodel -> obj` conversion path.

## What changed

- registered `.bbmodel` as a supported model input
- added a dedicated handler that parses Blockbench cube and mesh geometry
- exported parsed geometry as Wavefront `OBJ` using Three.js
- wired the handler into the existing handler registry
- added regression tests for cube and mesh export cases

## Notes

This resolves the current "unsupported `.bbmodel` input" gap with the smallest safe change in this repo's existing architecture.

The implementation targets geometry export to `OBJ`, which then allows reuse of existing downstream conversion paths. It does not add native `.blend` or `.mdl` export support.

## Testing

- added regression tests for cube-based `.bbmodel` export
- added regression tests for mesh-based `.bbmodel` export


neko782:
disclose ai usage next time and dont write tests unless there is *any* reason and unless they actually test for regressions

PR #626, opened 2026-04-16
fix: support Ogg Opus to Ogg Vorbis conversion

### LINKED ISSUE #603: [File format] OGG Opus -> OGG Vorbis
**Describe the file format**
Opus is an audio codec that is meant to replace the Vorbis codec. Currently, convert.to.it can convert OGG Vorbis files to OGG Opus files, but not the other way around.

FFmpeg should be able to do this in theory, but I'm not sure if a web build can.

No, I don't understand the FFmpeg handler well enough to do so.

**Additional context**
If FFmpeg isn't able to convert from Opus files, a web version of libopus (such as [this one](https://github.com/ImagicTheCat/libopusjs)) could be used instead.

valipuha:
## Summary
- add distinct FFmpeg aliases for Ogg Opus input and Ogg Vorbis output
- force `libvorbis` when converting to Ogg Vorbis
- add regression coverage for the new OGG codec-specific behavior

## Why
Previously the FFmpeg handler only exposed generic container-level `OGG`, so the conversion graph could not distinguish Ogg Opus input from Ogg Vorbis output even though both use `.ogg`.

## Testing
- targeted TypeScript compile passed
- targeted runtime helper validation passed

Closes #603

### DIFF
--- a/src/handlers/FFmpeg.ts
+++ b/src/handlers/FFmpeg.ts
@@ -7,6 +7,40 @@ import mime from "mime";
 import normalizeMimeType from "../normalizeMimeType.ts";
 import CommonFormats from "src/CommonFormats.ts";
 
+export function getSpecialAudioFormats(): FileFormat[] {
+  return [
+    {
+      name: "Ogg Opus Audio",
+      format: "opus",
+      extension: "ogg",
+      mime: CommonFormats.OGG.mime,
+      from: true,
+      to: false,
+      internal: "ogg",
+      category: "audio",
+      lossless: false
+    },
+    {
+      name: "Ogg Vorbis Audio",
+      format: "vorbis",
+      extension: "ogg",
+      mime: CommonFormats.OGG.mime,
+      from: false,
+      to: true,
+      internal: "ogg",
+      category: "audio",
+      lossless: false
+    }
+  ];
+}
+
+export function getCodecArgsForFormat(outputFormat: FileFormat): string[] {
+  if (outputFormat.internal === "ogg" && outputFormat.format === "vorbis") {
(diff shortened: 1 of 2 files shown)


src/handlers/FFmpeg.ts
```diff
@@ -1,13 +1,45 @@
...
 import { FFmpeg } from "@ffmpeg/ffmpeg";
 import type { LogEvent } from "@ffmpeg/ffmpeg";
 
 import mime from "mime";
 import normalizeMimeType from "../normalizeMimeType.ts";
-import CommonFormats, { Category } from "src/CommonFormats.ts";
-import { BadMagicError, EOFError, InitializationError } from "src/errors.ts";
+import CommonFormats from "src/CommonFormats.ts";
+
+export function getSpecialAudioFormats(): FileFormat[] {
```
neko782, line 10: .. a function that returns two formats?

src/handlers/FFmpeg.ts
```diff
@@ -1,13 +1,45 @@
...
+      from: false,
+      to: true,
+      internal: "ogg",
+      category: "audio",
+      lossless: false
+    }
+  ];
+}
+
+export function getCodecArgsForFormat(outputFormat: FileFormat): string[] {
```
neko782, line 37: a function with one if statement?!? also this is non conventional

test/handlers/ffmpeg.test.ts
neko782, line 1: these are useless

neko782 (commented):
disclose ai usage next time. i think ill just implement it myself

PR #644, opened 2026-04-27
Chippytech patch 1

chippytech:


chippytech:
s

maxjk0:
If you're still doing stuff on this, any chance you could get .md3 with the other 3d model formats? thats a big one that annoyingly only has like an obscure blender plugin to convert to, but also is still used in a few major modding communities

ConnorTippets:
one feature per pr please

PR #665, opened 2026-05-28
Add .mts (AVCHD) and .m2ts (Blu-ray) input support

EdmundLimBoEn:
## Summary

Adds FFmpeg input aliases for camcorder and Blu-ray MPEG-TS files:

- **`.mts` (AVCHD)** — FFmpeg auto-discovers `mpegts` but assigns the `.ts` extension, so AVCHD camcorder files (JVC, Sony, Panasonic) were unrecognised. Same pattern as `.wmv` → `asf`.
- **`.m2ts` (Blu-ray BDMV)** — same container, different extension used by Blu-ray rips.

Both entries are input-only (`to: false`). Converting *to* MPEG-TS remains available via the auto-discovered `.ts` muxer.

## Review feedback

Removed the AVCHD ZIP extractor. That handler existed only to paper over macOS silently zipping a folder when it is dragged into a browser. Handlers should not work around OS-specific upload behavior; users on any OS can unpack a ZIP with the existing archive handlers, then convert the `.mts` / `.m2ts` files directly.

## Test plan

- [ ] Upload a `.mts` file — recognised as “AVCHD Video” and convertible to MP4/WebM/etc.
- [ ] Upload a `.m2ts` file — same for “Blu-ray BDMV Video”
- [ ] `.ts` files still work (no regression on auto-discovered `mpegts`)
- [ ] Other FFmpeg formats unaffected

### DIFF
--- a/src/handlers/avchd.ts
+++ b/src/handlers/avchd.ts
@@ -1,0 +1,80 @@
+import type { FileData, FileFormat, FormatHandler } from "../FormatHandler.ts";
+import JSZip from "jszip";
+
+/**
+ * AVCHD ZIP handler.
+ *
+ * macOS silently zips an AVCHD folder when you drag it into a browser —
+ * the user ends up uploading a .zip instead of the individual .mts files
+ * inside BDMV/STREAM/.  This handler unpacks such a ZIP and returns every
+ * .mts / .m2ts file it finds, so the graph can route them through FFmpeg.
+ *
+ * Conversion path:  ZIP (avchd) → MTS  (then FFmpeg takes it to MP4/etc.)
+ */
+class avchdHandler implements FormatHandler {
+
+  public name: string = "AVCHD Extractor";
+  public supportedFormats: FileFormat[] = [];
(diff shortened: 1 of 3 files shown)


src/handlers/avchd.ts
```diff
@@ -0,0 +1,80 @@
+import type { FileData, FileFormat, FormatHandler } from "../FormatHandler.ts";
+import JSZip from "jszip";
+
+/**
+ * AVCHD ZIP handler.
+ *
+ * macOS silently zips an AVCHD folder when you drag it into a browser —
```
neko782: i dont think handlers should be working around weird os behavior

EdmundLimBoEn: Dropped the ZIP extractor. `.mts` / `.m2ts` are now just FFmpeg extension aliases (same pattern as `.wmv`), so they work the same on every OS. Existing archive handlers already cover unpacking if someone has a zipped AVCHD package.

neko782:
disclose ai usage next time

PR #687, opened 2026-06-12
Added Support For kra & krz (Krita Files) to PNG

### LINKED ISSUE #684: [File format] .kra - krita documents
**Describe the file format**
.kra is the extension for krita files, documents. They're used in the krita app. They're images. Well rich images at least because they got the layers and stuff yk

Support it like you would with any image, squash the layers into one when converting to stuff that doesn't support multiple layers tho

Yeah that idk tho maybe the krita app itself ig

**PLEASE NOTE:** Simply listing a file format with no further justification beyond "I want this" will likely get your issue closed. There are a lot of file formats out there, and it can take hours to add support for just one. Just naming them isn't productive, so please try to be forthcoming with resources of your own.

Ok well I gave some info hope this is enough.

No

**Additional context**
No further context.

AFAskar:
resolves #684 

This PR only implements the conversion from .kra & .krz to .png files, not the reverse. While the reverse conversion is possible, I have not had enough time to implement it. However, the code within the Krita codebase that handles that functionality can be found
[here](https://invent.kde.org/graphics/krita/-/blob/master/libs/ui/kis_png_converter.cpp?ref_type=heads)

i attempted to do it using LLMs for assistance (my cpp is very lacking), but i wasn't able to reach a good general solution so i didn't commit that attempt

the kra to png is fully my code (using the comic.ts as reference since they are both archive based formats)


src/handlers/kra.ts
```diff
@@ -0,0 +1,69 @@
...
+
+class kraHandler implements FormatHandler {
+  public name: string = "kra";
+  public supportedFormats?: FileFormat[];
+  public ready: boolean = false;
+
+  async init() {
+    this.supportedFormats = [
+      CommonFormats.PNG.builder("png")
+        .markLossless()
```
ConnorTippets: this conversion isn't lossless, cause it merges all layers into one

src/handlers/kra.ts
```diff
@@ -0,0 +1,69 @@
...
+      {
+        name: "Krita Raster Archive (KRA)",
+        format: "kra",
+        extension: "kra",
+        mime: "application/x-krita",
+        from: true,
+        to: false,
+        internal: "kra",
+        category: ["archive"],
+        lossless: true,
```
ConnorTippets: ditto

src/handlers/krz.ts
```diff
@@ -0,0 +1,60 @@
...
+
+class krzHandler implements FormatHandler {
+  public name: string = "krz";
+  public supportedFormats?: FileFormat[];
+  public ready: boolean = false;
+
+  async init() {
+    this.supportedFormats = [
+      CommonFormats.PNG.builder("png")
+        .markLossless()
```
ConnorTippets: ditto.

src/handlers/krz.ts
```diff
@@ -0,0 +1,60 @@
...
+      {
+        name: "Krita Raster Archive (krz)",
+        format: "krz",
+        extension: "krz",
+        mime: "application/x-krita",
+        from: true,
+        to: false,
+        internal: "krz",
+        category: ["archive"],
+        lossless: true,
```
ConnorTippets: ditto.

ConnorTippets (changes requested):

AFAskar:
It's been done, but a question
It will be considered lossless if each layer was exported as an image ?
@ConnorTippets

ConnorTippets:
> It's been done, but a question It will be considered lossless if each layer was exported as an image ? @ConnorTippets

yeah, pretty much. you might lose some exact color calibration settings but it would be close enough

neko782 (approved):

PR #688, opened 2026-06-14
txt to pdf implementation

Stijn7683:
It uses PDFKit to convert text to a pdf file. Before converting it removes any emojis, because otherwise emojis would end up like 'Ø=ÜÄ'. If the file doesn't contain any text but it does contain emojis then it throws an error.

The text gets placed into A4 sized page(s) with a margin of 72 on all the sides. The text has the font 'Courier'.

I used AI to make the conversion code.


p2r3:
This does not build. You can check with `bunx vite build` I think.

Stijn7683:
> This does not build. You can check with `bunx vite build` I think.

The build completed successfully for me. 
There was an import error and I asked ai to fix it. Hopefully it works now.

PR #695, opened 2026-06-16
buildCache.js: Get next available port if 8080 is in use

exurd:
SISR needs to use Steam's CEF debugging mode, which always uses port 8080. That breaks the buildCache.js, not good. This PR fixes that by finding the nearest available port for the server.

### DIFF
--- a/buildCache.js
+++ b/buildCache.js
@@ -9,15 +9,27 @@ if (await outputFile.exists()) {
   await outputFile.delete();
 }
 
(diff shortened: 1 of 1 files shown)


neko782:
https://github.com/p2r3/convert/commit/e0df21692650a0d29dee9266a315269a675e086b

exurd:
That is some clean looking code. Well done!

PR #704, opened 2026-07-10
(plain) BITMAPINFOHEADER (to) support

### LINKED ISSUE #705: [Feature] V1 BITMAPINFOHEADER
**Describe the feature**
The original BITMAPINFOHEADER is one of the most common bitmap headers, likely due to its combination of simplicity, versatility, and every header after it being much like a version with more stuff. Additionally it is very good for learning how bitmaps work, likely due to, once again, its simplicity.

It can (entirely losslessly) convert from (many) other image formats and from raw rgb/rgba data, but the main conversion happens from the latter. regardless of the starting point, the finish point is a BITMAPINFOHEADER-headed BMP file. if converting from an image, it will reuse the dimensions, bit depth, and bit layout of that image, and if not, it will calculate the first two manually and use preset options for the second.

I'm not sure of any existing implementations that specifically use BITMAPINFOHEADER (if i haven't mentioned it already, the header that appears to be used here is BITMAPINFOV5HEADER). references on the format (in general) are available [here](https://en.wikipedia.org/wiki/BMP_file_format#File_structure) and [here](https://www.fileformat.info/format/bmp/egff.htm), with the first one being my main source of information on how this specific header works.

#704 

**Additional context**
I did a bit of a search on issues before making this one and found that #148 could be related if I am able to add support for a custom palette in addition to the default one i currently have set up (as well as more compression options i suppose). letting @cs-deathmatch know.

bas1licate:
I may put a more thorough explanation in the issue but in "short", (plain 40-byte) BITMAPINFOHEADER is generally worth adding in my opinion. first, according to [en-wikipedia](https://w.wiki/SEKA) [for at least 15 years](https://w.wiki/SEKB), this version is a common header. second, it is also very nice for learning how the bmp format works, so it has that value as well. for these reasons, i think supporting it in addition to the already-supported BITMAPINFOV5HEADER used for most if not all other conversions to BMP would be a good idea.

~~lastly, i would perhaps like to dedicate this pr to @omega-fallon for making rgba.ts, and also for being a super contributor to the repo in general, but honestly, who dedicates PRs?~~

(may) Resolve #705


omega-fallon:
There's a first time for everything! Thank you for the shout-out :)

bas1licate:
now the reason this is staying as draft is because the code still contains a few issues i have yet to fix.

bas1licate:
all things considered this doesn't seem to be going anywhere. will just close for the time being.

PR #707, opened 2026-07-21
optimize

tutududu2:


neko782:
removed main.ts changes because they are incompatible with the new ui

PR #715, opened 2026-08-12
Infinite craft to JSON

Rubiksmath:
friend here
i was debating ic->txt but decided against it because lossless
i put info about ic file in the typescript
can't PR myself sorry, personal issues with github account
follows up from #299
let me know if there are issues thanks i'll read this PR


src/handlers/index.ts
```diff
@@ -38,7 +38,7 @@ import turbowarpHandler from "./turbowarp.ts";
 import libopenmptHandler from "./libopenmpt.ts";
 import { midiCodecHandler, midiSynthHandler } from "./midi.ts";
 import lzhHandler from "./lzh.ts";
-import txtToInfiniteCraftHandler from "./txtToInfiniteCraft.ts";
+import { txtToInfiniteCraftHandler, infiniteCraftToJsonHandler } from "./infiniteCraft.ts";
```
neko782, line 41: should probably be a single handler

neko782:
oh it is for avoiding bad paths. sure, i guess

PR #723, opened 2026-08-23
WAVEBREAKER

bas1licate:
the waves break.

this is sort of a prelude to bminfo, both being about generating formats which support raw data, except this one would be much more appealing to me due to any potential padding being minimal instead of somewhat complex. i couldn't find a raw audio data/samples format though, so until one is found or created this just temporarily supports any input. i should mention this should probably also only be used in advanced mode but i don't really think there's a way to make that happen just yet.

was tested [as standalone](https://bas1licate.github.io/wavebreak) and it does work, so if it doesn't work here it's probably just because #73 finally happened while i was working on it or something like that.

also i may or may not dedicate this pr to the calendar date of 1993 january 7, which has [no particular significance whatsoever](https://drive.google.com/file/d/1ReskLSrvJW3S72RWpghaN1MWHPzbY_K0/view).


bas1licate:
information which may be important: ffmpeg has pcm audio types which is exactly what i'm looking for (in my case i (currently) "need" the `s16le` type) so if that could be handled somehow, either now or in the future, that would be great

p2r3:
Could you give some information on what this pull request actually does? It's not linked to an issue, and the description is incredibly vague. I shouldn't have to dig through the code to understand what you're here to do.

bas1licate:
i suppose i completely forgot to explain what it does (whoops) so let me explain it now:
so basically what it does is it takes data and makes a WAV using that data as 16-bit PCM samples. that's it.

didn't know of any sample formats when making it and even know when i know some i'm not sure how to specify the ffmpeg `s16le` format as the input format, much less make a way to convert something to that in the first place, so because of that it currently just accepts any input.

bas1licate:
also for issue stuff i did technically in a way mention this being conceptually related to #704 (which is still unfinished until i find a way to fix whatever's wrong with it), wav being the auditory equivalent of bmp and all, except as i said even more close to raw data than bmp is, and thusly much easier to make a thing for.

bas1licate:
_Originally posted by @p2r3 in https://github.com/p2r3/convert/issues/723#issuecomment-5384653979_
> Could you give some information on what this pull request actually does? It's not linked to an issue, and the description is incredibly vague. I shouldn't have to dig through the code to understand what you're here to do.

once again thank you for having reminded me to do that and so I had therefore done it not long after I saw you saying that

neko782:
smells too much like a binary waterfall, you shouldnt declare that you can convert from anything if you can actually only convert from 16 bit raw samples

neko782 (changes requested):
forgor to add it to index.ts!

bas1licate:
i'm just gonna put this on the backburner until either I know how to use ffmpeg types or someone else does and uses that knowledge here

bas1licate:
just recently learned about mime audio/L16 i'll be using that

neko782:
you still gotta add it to index.ts

bas1licate:
k will do before undrafting

neko782 (approved):

neko782:
great job

PR #725, opened 2026-08-26
fix: resolve ambiguous format tokens by candidate, not by handler order

ogfrench:
## What

The agent surfaces (MCP `convert_file` / `find_conversion_path`, REST `POST /convert` / `GET /path`) resolve a `(mime, token)` pair rather than being handed the `FileFormat` a user clicked, so they have to guess which registry entry a token means. `json` matches nineteen entries, `png` thirty.

The guess was "first match in handler order". Measured against the real Node handler set:

| pair | before | after |
|---|---|---|
| `csv→json` | `libreoffice:csv → pandoc:csljson` (emits CSL) | `libreoffice:csv → tojson:json` |
| `json→csv` | `pandoc:csljson → pandoc:json → fromjson:csv` | `pandoc:json → fromjson:csv` |
| `md→pdf` | **no path** | `pandoc:commonmark → pandoc:html → libreoffice:pdf` |
| `epub→pdf` | **no path** | `pandoc:epub → pandoc:html → libreoffice:pdf` |
| `pdf→txt` | `Ghostscript → tiff → png → html → pandoc` | `Ghostscript:pdf → pdfparse:text` |
| `docx→pdf` | `libreoffice → zip → png → imageToPdf` | `libreoffice:docx → libreoffice:pdf` |

`md→pdf`, `epub→pdf` and `png→svg` all have landing pages advertising them.

## How

`findFormatCandidates` ranks the entries a token could mean — an exact `format` match above one that only matched `extension`, ties keeping handler order. `findFirstPath` tries the pairings in that order rather than committing to the first, because a token can resolve to a perfectly good reader that is still a dead end for this particular target. Capped at 24 pairings so a token with thirty readers cannot hang a request.

This is the strategy `scripts/verify-conversions.ts` has always used to prove the pairs work. It now lives where the shipped code can use it.

Two things fell out of putting all four call sites on one helper:

- **MCP was using the handler-constrained search mode**, REST the simple one. The graph keys nodes by format, not handler, so naming a handler on the target bought no precision — it only rejected paths whose last hop came from a different handler writing the same format, leaving whatever longer route survived. That is the `pdf→txt` and `docx→pdf` rows above.
- **An unknown format was turned away without asking the browser bridge**, on the reasoning that the bridge could not resolve it either. Backwards: the bridge is the side with the browser-only handlers. `png→svg` needs `svgTrace`, which never loads under Node.

## Also

`pdfparse` handed pdf.js the caller's own `ArrayBuffer` — the only one of five pdf.js call sites in the repo not to copy. pdf.js detaches it, so any second use threw `Underlying ArrayBuffer has been detached from the view`: a multi-hop route reusing the bytes, `convert_file` falling back to the bridge after a native failure, or one sample PDF converted to several targets in a row. Latent until 3.0.0's worker-resolve fix made this handler reachable under Node at all.

## Why CI changes

Not one of these would have failed a route-graph assertion; every one was found by running the conversion. `scripts/verify-conversions.ts` goes in as its own job with LibreOffice installed, so the office pairs are exercised rather than skipped. ~2.5 minutes, running beside `build` rather than after it.

## Verification

- `bun run test` — 1249 pass. (`src/mcp/index.integration.test.ts` times out under parallel load; passes on its own, as before.)
- `bun run scripts/verify-conversions.ts` — **52 PASS / 0 FAIL** (was 51 PASS / 1 FAIL).
- `bun x tsc` clean, `bun x knip --include files,unlisted` clean.
- 8 new resolver tests; 4 existing tests updated to the corrected unknown-format behaviour.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ogfrench:
Opened against the wrong repository by mistake; this belongs on ogfrench/frogConvert. Closing.

PR #728, opened 2026-08-31
Fix #500: [Bug] PNG <-> TXT

### LINKED ISSUE #500: [Bug] PNG <-> TXT
When you convert from a PNG to a TXT, it converts it in a block of text based on its light intensity.. So the resolution gets reduced by a lot and it turns grayscale. Converting it back to an image just takes a screenshot of it (as you mentioned in the video).

1. Input png and convert to txt
2. Convert txt back to png

**Expected behavior**
It should not reduce the resolution of the image when converting to a txt. And when converting back, instead of taking a screenshot, it can take the individual character, and based on the defined intensity order, it could create the pixel itself (grayscale).
For getting color data, we could use ANSI color format code for every character, though that wouldn't display the ascii art correctly. I'm not sure myself of what else can be done to preserve the color data.

**Screenshots**

Image
Image
Image

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser Chrome
 - Version 145.0.7632.117 (Official Build) (64-bit)

**Additional context**
If you want, you can also take a look at my project on converting image and video (colored and grayscale) to ascii art and animation in command prompt.
https://github.com/MagnaMadness/Video_to_ASCII

victorwon2001:
Fixes #500

## Implementation summary

PNG → TXT now writes a full-resolution grayscale ASCII grid with exactly one palette character per source pixel. TXT → image recognizes a non-empty rectangular grid made entirely from that palette, infers width from the row length and height from the number of rows, and reconstructs opaque grayscale pixels at the same spatial resolution. Text that does not fully parse as a rectangular palette grid continues through the existing text-rendering path. As discussed in review, rectangular palette-only TXT is intentionally ambiguous and is treated as image data without a magic tag or explicit dimension metadata.

## Changes

- Removed the preview, start/end markers, and `width=`/`height=` metadata from image → TXT output.
- Preserved one character per original image pixel instead of reducing the image to 16×24-cell preview resolution.
- Infer TXT image dimensions directly from equal-length rows and the number of rows.
- Accept CRLF input and one ordinary final line ending.
- Fall back to the existing text screenshot behavior for empty text, ragged rows, or text containing characters outside the intensity palette.
- Kept grayscale palette quantization, safe dimension/allocation checks, and `ArrayBuffer`-backed `Uint8ClampedArray` output for `ImageData` compatibility.
- Moved the codec helper under `src/handlers/canvasToBlob/` and its detailed regression tests to `test/handlers/canvasToBlob.test.ts`.
- Kept supported formats, output naming, ordinary TXT rendering, the non-TXT image path, and the conversion graph unchanged.

## Testing

- Focused source and diff validation completed.


neko782:
isnt it better to preserve being able to actually see the image when we convert it to text

neko782:
nvm you can see it. no need for the preview

src/handlers/canvasToBlob/imageTextCodec.ts
neko782, line 1: there should not be random helpers in /handlers put it in a subfolder

test/commonFormats.test.ts
```diff
@@ -105,6 +106,82 @@ test("png → svg", async () => {
 
 }, { timeout: 60000 });
 
+test("png → tagged txt → png preserves dimensions and grayscale pixels", async () => {
```
neko782: this test should not be here

src/handlers/imageTextCodec.ts
```diff
@@ -0,0 +1,183 @@
+export const IMAGE_TEXT_MARKER = "␞ CONVERT.TO.IT IMAGE TEXT V1 ␞";
```
neko782: do we have to tag it and give it width and height? i mean, dimensions are implicit by amount of pixels, and if something parses as ascii art it might as well be

neko782 (changes requested):
disclose ai usage next time

victorwon2001:
Removed the image-text tag and explicit width/height metadata. TXT output is now only the full-resolution rectangular palette grid, and TXT-to-image infers dimensions directly from its rows and columns; rectangular palette-only user text is intentionally treated as an image. The codec and focused tests were also moved to canvasToBlob-specific locations, with focused tests, production build/cache generation, and all 9 broad conversion tests passing.

src/handlers/canvasToBlob.ts
```diff
@@ -98,13 +114,20 @@ class canvasToBlobHandler implements FormatHandler {
...
+          grayscale[i] = rgbaToGrayscale(
+            pixels.data[index],
+            pixels.data[index + 1],
+            pixels.data[index + 2],
+            pixels.data[index + 3],
+          );
+        }
+        bytes = new TextEncoder().encode(encodeImageText({
+          width: pixels.width,
+          height: pixels.height,
```
neko782, line 133: there gotta be some downscaling at some point (512? 256?), a 6mb text file is useless

neko782:
im not sure on this one, it works and achieves the goal but we lose a high effort encoder for small previews (which could be what the user wants?) i also dont appreciate non disclosed ai

victorwon2001:
Removed the orphan `image-to-txt` gitlink from the candidate while keeping the obsolete submodule deleted, and the focused checks plus the full Pages build/cache/test sequence now pass in a normal recursive-submodule checkout. AI assistance was used for this follow-up.

PR body correction:

Fixes #500

## Implementation summary

PNG → TXT now writes a grayscale ASCII grid with exactly one palette character per resulting pixel. Images whose width and height are both at most 512px keep their original dimensions; larger images are downscaled so the longest side is 512px while preserving aspect ratio. TXT → image recognizes a non-empty rectangular grid made entirely from that palette, infers width from the row length and height from the number of rows, and reconstructs opaque grayscale pixels at the grid's spatial resolution. Text that does not fully parse as a rectangular palette grid continues through the existing text-rendering path. As discussed in review, rectangular palette-only TXT is intentionally ambiguous and is treated as image data without a magic tag or explicit dimension metadata.

## Changes

- Removed the preview, start/end markers, and `width=`/`height=` metadata from image → TXT output.
- Preserve one character per resulting image pixel; images with both dimensions at or below 512px retain their original dimensions, while larger images are downscaled to a 512px longest side with aspect ratio preserved.
- Infer TXT image dimensions directly from equal-length rows and the number of rows.
- Accept CRLF input and one ordinary final line ending.
- Fall back to the existing text screenshot behavior for empty text, ragged rows, or text containing characters outside the intensity palette.
- Kept grayscale palette quantization, safe dimension/allocation checks, and `ArrayBuffer`-backed `Uint8ClampedArray` output for `ImageData` compatibility.
- Moved the codec helper under `src/handlers/canvasToBlob/` and its detailed regression tests to `test/handlers/canvasToBlob.test.ts`.
- Kept supported formats, output naming, ordinary TXT rendering, the non-TXT image path, and the conversion graph unchanged.

## Testing

- Focused source and diff validation completed.

neko782:
we are not accepting contributions from automated entities at this time. sorry for wasting your tokens, i wasnt sure about outright closing this

PR #735, opened 2026-09-09
Html to exe

TKTK123456:
Adds support for converting html to exe, used AI for getting the file that it is ran with and for the general method of execution. I looked through the code and edited things so that it would work how I wanted it to.


neko782 (approved):

neko782:
can i have some sources for the binary before merging?

neko782:
@TKTK123456

TKTK123456:
When I get home I will try

TKTK123456:
[stub.zip](https://github.com/user-attachments/files/32205349/stub.zip)
This was the code used to generate it

TKTK123456:
@neko782

neko782:
ill rebuild it from source just in case and then merge

neko782:
does it run on native windows? 

edit: ill test it myself

neko782 (changes requested):
takes like 15 seconds before it loads, opens a spurious cmd and doesnt close (except from the cmd?..)

PR #738, opened 2026-09-10
Fix example handler typo

BunnyNabbit:
The example handler contained a typo where one instance of `Category` was spelt out as `CATEGORY`, which points to an unused identifier. This change fixes that typo, so pasting it in a new file won't result in issues related to that.


neko782:
good one

PR #741, opened 2026-09-11
refactor like, everything

neko782:
the one million line patch. the one and only. the ultimate pull request. which also breaks all your branches

KINDLY REQUESTING to not start any big work before this is merged (for your and my sanity). if anyone was doing a big ahh pull request privately, nows the time to either make it (which i will pull) or notify me here to hold off while you are still making it. if its like, one file of additions, you arent affected


neko782:
everyone, start big new work IMMEDIATELY. i need pull requests to merge

